PR #127, opened 2018-08-27
frontend: greet user on dashboard

prtksxna:


prtksxna:
@slaporte This is the simplest change I could see from Margeigh's slides. Just wanted to get the ball rolling 😃

PR #149, opened 2019-05-24
2019 Wikimedia Hackathon Service Pack

yarl:
I will add list of issues closed soon.

### DIFF
--- a/client/src/components/vote/vote-single/vote-single.html
+++ b/client/src/components/vote/vote-single/vote-single.html
@@ -1,14 +1,26 @@
-<div class="image"
-   layout="row" layout-align="center stretch"
-   ng-if="!$ctrl.error && $ctrl.round.status === 'active' && $ctrl.images.length">
-
-  <div class="image__container" flex="none"
-      layout="row" layout-align="center center" 
-      ng-class="$ctrl.showSidebar ? 'image__container--with-sidebar' : 'image__container--no-sidebar'"
-      mont-key-actions actions="$ctrl.keyDown">
-    <img mont-src="{{ $ctrl.getImageName($ctrl.rating.current) }}">
-    <md-button class="md-icon-button image__sidebar-hide-btn"
-        ng-click="$ctrl.showSidebar = !$ctrl.showSidebar">
+<div
+  class="image"
+  layout="row"
+  layout-align="center stretch"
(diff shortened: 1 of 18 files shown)


atsirlin:
@yarl: Thanks! I am curious if there were any changes in the back-end after the hackathon, or the 'close campaign' button has been implemented without the actual possibility of closing the campaign.

mahmoud:
@atsirlin no backend changes (from me) yet! I'll review this and then reply your email, @atsirlin, but while I'm here I might as well say thank you for working on this at the hackathon. I was planning on doing more, but I caught a bad cold. Glad you two still got a bunch done, great work! OK reviewing now...

atsirlin:
@mahmoud: thank you too, it was a pleasure to meet Pawel and get a bit more familiar with the code! However, it would also be great if someone replies to the e-mail that I sent around after the hackathon...

client/src/components/vote/vote-single/vote-single.html
```diff
@@ -138,33 +167,43 @@ <h3>
...
             </p>
           </div>
-          <md-button class="md-secondary md-icon-button"
-              ng-href="{{ '//commons.wikimedia.org/wiki/File:' + $ctrl.rating.current.entry.name + '#filehistory' }}"
-              target="_blank">
+          <md-button
+            class="md-secondary md-icon-button"
+            ng-href="{{
+              '//commons.wikimedia.org/wiki/File:' +
+                $ctrl.rating.current.entry.name +
```
mahmoud: is this going to need escaping or is `entry.name` already url escaped?

slaporte:
This is looking good to me.

client/src/components/vote/vote-single/vote-single.html
slaporte: I think we're getting `entry.name` straight from mediawiki, so it should be URL friendly. 

We should have a test for this, though—`$ctrl.rating.current.entry.name` makes an appearance above, and there is `encodeURIComponent(image.entry.name)` elsewhere...

slaporte:
> However, it would also be great if someone replies to the e-mail that I sent around after the hackathon...

@atsirlin could you re-send, if those questions are still unanswered?

atsirlin:
I think it's fine for now

Issue #115, opened 2017-11-28
Provide statistics feedback to voters


dschwen:
Some contests suggest a curve grading for images (in particular in 5star mode, but maybe also in yes/no mode). It would be nice to see a bar graph (or even just percentage numbers) for the amount of the different ratings given by the current user.

slaporte:
I like this idea. 

Are you interested in seeing your own rating stats, as a juror, or seeing the full jury's stats, as the campaign coordinator? Both?

dschwen:
I think it would be unproblematic to show the juror's own voting stats. Jurywide stats could be an optional per round setting.

MichaelNMaggs:
This would be very useful, as in WLM we often want to tell reviewers that they should select roughly, say, one in five images for the next round.

PR #153, opened 2019-09-02
Vote statistics prototype

baturin:
The pull request is the first step to resolve https://github.com/hatnote/montage/issues/115. It adds count of accepted/declined images to "edit votes" page of juror, with ability to enable/disable statistics through "edit round" page (statistics is disabled by default).

### DIFF
--- a/montage/juror_endpoints.py
+++ b/montage/juror_endpoints.py
@@ -37,6 +37,7 @@ def get_juror_routes():
            POST('/juror/round/<round_id:int>/tasks/submit', submit_ratings),
            POST('/juror/round/<round_id:int>/tasks/skip', skip_rating),
            GET('/juror/round/<round_id:int>/votes', get_votes_from_round),
+           GET('/juror/round/<round_id:int>/votes-stats', get_votes_stats_from_round),
            GET('/juror/round/<round_id:int>/ratings', get_ratings_from_round),
            GET('/juror/round/<round_id:int>/rankings', get_rankings_from_round),
            POST('/juror/round/<round_id:int>/<entry_id:int>/fave', submit_fave),
@@ -63,7 +64,8 @@ def make_juror_round_details(rnd, rnd_stats, ballot):
            'total_open_tasks': rnd_stats['total_open_tasks'],
            'percent_tasks_open': rnd_stats['percent_tasks_open'],
            'ballot': ballot,
-           'campaign': rnd.campaign.to_info_dict()}
+           'campaign': rnd.campaign.to_info_dict(),
+           'show_stats': rnd.show_stats,}
     return ret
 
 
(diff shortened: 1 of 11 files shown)


montage/juror_endpoints.py
```diff
@@ -152,6 +154,24 @@ def get_votes_from_round(user_dao, round_id, request, rnd=None):
...
+
+    if rnd.vote_method == 'yesno':
+        raw_ratings = juror_dao.get_rating_stats_from_round(round_id)
+        stats = {'yes': 0, 'no': 0}
+        vote_map = {0.0: 'no', 1.0: 'yes'}
+        for count, rating_value in raw_ratings:
+            if rating_value in vote_map:
+                stats[vote_map[rating_value]] = count
+
+        result = {'method': 'yesno', 'stats': stats}
```
mahmoud: would you like to extend this to support star-rating rounds, too? Also, what should the behavior be in a ranking round? Maybe just "ranked" and "unranked" counts? (unranked being the default lowest value an entry can get in a ranked round)

montage/rdb.py
```diff
@@ -276,6 +276,7 @@ class Round(Base):
     status = Column(String(255), index=True)
     vote_method = Column(String(255))
     quorum = Column(Integer)
+    statistics = Column(Boolean)
```
mahmoud: Is there a case where we'd want to prevent statistics from being sent down? If so, it may be easier to simply put a `show_stats` flag in the `flags` column. Then we can kick the can down the road on a migration.

montage/juror_endpoints.py
atsirlin: This feature is mostly needed in those cases where we set up limits for the number of 'yes' votes, 5-star votes, etc. It does not apply to the ranking round, where all photos are just ranked anyway.

atsirlin:
Great work! Let me mention that we would highly appreciate to have this deployed within the next week, so that we could use this feature during WLM.

montage/rdb.py
baturin: Fixed: store `show_stats` in the flags `column`

montage/juror_endpoints.py
baturin: 1. Extended support for star-rating rounds.
2. Avoid show stats checkbox when editing ranking rounds.

baturin:
> Impressive work! I've been meaning to cook up a contributor guide for a while, for exactly this sort of PR, but here you go, proving I don't need to! (joking, a guide is still top priority for me, will be my next PR, I swear).
> 
> I added some comments inline. I also think we may want to extend the integration test to at least touch these features, too. Excellent progress!
Thanks for your comments!
Where can I find the integration tests?

mahmoud:
Hey again @baturin, thanks for your patience with my reply.  Tests are usually just added to this long integration test here: https://github.com/hatnote/montage/blob/master/run_server_test.py

The test reads a bit like a script of how admins and jurors use Montage, so take a minute to familiarize yourself with the whole flow so that you can decide where to add this feature's tests (e.g., one toward the beginning of round creation, and another later on).

The test is run by having a development server running in one console, and then executing `python run_server_test.py` in another.

Thanks again!

atsirlin:
@mahmoud: what are the chances of putting this change into production? We are about to start grading for WLM and would be really grateful for having the vote statistics available.

mahmoud:
@atsirlin I'd be happy to deploy, it would be good to have a test to make sure it works (and continues to work), but if @baturin doesn't have time, we can defer that and deploy sooner. When does the actual judging start?

atsirlin:
@mahmoud: we plan to start this week.

atsirlin:
Can I be of any help with the testing? Is this feature already available on montage-dev?

mahmoud:
@atsirlin This is for the automated test, as detailed in my comment. Is it this week like tomorrow or this week like Thursday/Friday? I think I can sneak a test of my own in there if @baturin doesn't have the time atm. I may need to ping @yarl to do a frontend release.

atsirlin:
@mahmoud: thanks for the prompt response! We may still wait till Thursday or Friday. Now I am basically trying to understand if it makes sense to wait, or we better start now, if there is little chance to get the new feature in the production soon.

run_server_test.py
```diff
@@ -399,6 +399,11 @@ def full_run(base_url, remote):
                  {'post': True},
                  as_user='LilyOfTheWest')
 
+
+    resp = fetch('juror: get votes stats',
+                 '/juror/round/%s/votes-stats' % round_id,
+                 as_user='Slaporte')
```
mahmoud: did you want to check the response format at all? Just so it doesn't diverge too much from what the angular expects?

baturin:
@mahmoud First of all, thank you for the assistance!
Unfortunately it seems I won't be able to create adequate tests this week, I don't have enough time at the moment. I've finally managed to start them, but the full run takes more than 1 hour on my environment, which makes me think that I should optimize the environment somehow first...

mahmoud:
@baturin Whoa! These tests run in about 2 minutes for me, using a sqlite database. But I say we can always circle back and improve them later. Thanks for the great work!

mahmoud:
(will ping back here once it's deployed, hopefully today or tomorrow)

atsirlin:
For the information, we could not wait any longer and started the grading for Russia tonight. Nevertheless, it is highly desirable to have this feature deployed ASAP.

mahmoud:
Took longer than expected to get the frontend deployed (not my are of expertise until *very* recently ;) ), but it's deployed now!

Screenshot from 2019-09-22 16-58-33

We moved the stats over to the right, but they're there for yes/no and star rating rounds! Thanks again!

PR #159, opened 2020-02-09
Improve votes stats test assertions

baturin:
Improved votes stats test assertions which were added at https://github.com/hatnote/montage/commit/9b30bafdbbac47d57815f8bdfd3ddbcd7e4388a2


mahmoud:
More tests! Excellent! Would you mind merging in master so we can let CI run over it? Thanks!

mahmoud:
I took the liberty of updating the source branch. Good news is that CI looks green, so I'll be merging. Thanks again!

Issue #157, opened 2019-10-19
Not closed campaigns are showed under Your Closed Campaigns


ferdi2005:
Hi,
also not closed campaigns are showed under Your Closed Campaigns.
(All campaigns are showed there).

mahmoud:
Hi @ferdi2005. We've refactored the listing into an Active and All, so I'm going to close this for now.

Issue #144, opened 2019-05-18
Hide closed campaigns in the juror view


atsirlin:
-- Back-end --
In the juror view, provide an endpoint with the information on whether a campaign has been closed or not.

-- Front-end --
Place closed campaigns in the end of the list and make them wrapped by default.

MichaelNMaggs:
Same as #122

slaporte:
We've deployed the archive feature.

Issue #122, opened 2018-03-16
Hide details of old campaigns from jurors


MichaelNMaggs:
When logging in, jurors that have used the tool previously are presented with a list of all old (completed) campaigns before the one that's currently live. I for example am shown 18 old entries, which means I have to scroll down 9 screens to get to the one that's live, right at the bottom.    Many of those I know nothing about (eg specific country-based campaigns for WLAfrica), even though I'm listed as having completed 100% against each one.

Would prefer to have all the old stuff hidden, or at least have the current campaign presented first.

slaporte:
We just deployed a new version with an archive button to hide old campaigns. Thanks to @baturin and @mahmoud for this feature!

We auto-archived all the campaigns that closed in 2019. Coordinators can archive any other campaigns that are done, and can un-archive their campaigns if needed.

PR #169, opened 2020-06-18
Force finalize campaign

baturin:
That pull request is aimed to start discussion on a problem when admin sees a lot of open campaigns which are not actual anymore.

Simple proposed solution in that PR is to allow closing campaigns regardless of their current state and active rounds. Could you please clarify what could be wrong with that approach?

### DIFF
--- a/client/src/components/campaign/campaign.component.js
+++ b/client/src/components/campaign/campaign.component.js
@@ -34,6 +34,8 @@ function controller($filter, $q, $state, $stateParams, adminService, alertServic
         if (!vm.campaign.rounds.length) {
           vm.campaign.rounds.push({});
         }
+
+        vm.isCampaignClosed = vm.campaign.status === 'finalized';
       })
       .catch((err) => {
         vm.err = err;
--- a/client/src/components/campaign/campaign.component.js
+++ b/client/src/components/campaign/campaign.component.js
@@ -1,5 +1,6 @@
 import './campaign.scss';
 import template from './campaign.html';
+import { CAMPAIGN_STATUS_ACTIVE, CAMPAIGN_STATUS_FINALIZED } from "../../constants";
 
 const Component = {
   controller,
@@ -36,7 +37,8 @@ function controller($filter, $q, $state, $stateParams, adminService, alertServic
           vm.campaign.rounds.push({});
         }
 
-        vm.isCampaignClosed = vm.campaign.status === 'finalized';
+        vm.canCloseCampaign = vm.campaign.status === CAMPAIGN_STATUS_ACTIVE;
+        vm.canReopenCampaign = vm.campaign.status === CAMPAIGN_STATUS_FINALIZED;
       })
       .catch((err) => {
         vm.err = err;
--- a/client/src/components/campaign/campaign.component.js
+++ b/client/src/components/campaign/campaign.component.js
@@ -15,6 +15,7 @@ function controller($filter, $q, $state, $stateParams, adminService, alertServic
   vm.newRound = false;
 
   vm.closeCampaign = closeCampaign;
+  vm.reopenCampaign = reopenCampaign;
   vm.editCampaign = editCampaign;
   vm.saveEditCampaign = saveEditCampaign;
 
@@ -70,6 +71,17 @@ function controller($filter, $q, $state, $stateParams, adminService, alertServic
     vm.loading = true;
     adminService
       .finalizeCampaign(vm.campaign.id)
+      .then(() => $state.reload())
+      .catch(alertService.error)
+      .finally(() => {
+        vm.loading = false;
+      });
+  }
+
+  function reopenCampaign() {
+    vm.loading = true;
+    adminService
+      .reopenCampaign(vm.campaign.id)
       .then(() => $state.reload())
       .catch(alertService.error)
       .finally(() => {
--- a/client/src/components/campaign/campaign.html
+++ b/client/src/components/campaign/campaign.html
@@ -42,6 +42,14 @@
         Close Campaign
       </md-button>
       <md-button
+        aria-label="Reopen Campaign"
+        ng-if="!$ctrl.edit && $ctrl.isCampaignClosed"
+        ng-click="$ctrl.reopenCampaign()"
+      >
+        <md-icon>assignment_turned_in</md-icon>
+        Reopen Campaign
+      </md-button>
+      <md-button
         aria-label="Edit Campaign"
         ng-if="!$ctrl.edit"
         ng-click="$ctrl.editCampaign(true)"
--- a/client/src/components/campaign/campaign.html
+++ b/client/src/components/campaign/campaign.html
@@ -35,7 +35,7 @@
       </md-button>
       <md-button
         aria-label="Close Campaign"
-        ng-if="!$ctrl.edit && !$ctrl.isCampaignClosed"
+        ng-if="!$ctrl.edit && $ctrl.canCloseCampaign"
         ng-click="$ctrl.closeCampaign()"
       >
         <md-icon>assignment_turned_in</md-icon>
@@ -43,7 +43,7 @@
       </md-button>
       <md-button
         aria-label="Reopen Campaign"
-        ng-if="!$ctrl.edit && $ctrl.isCampaignClosed"
+        ng-if="!$ctrl.edit && $ctrl.canReopenCampaign"
         ng-click="$ctrl.reopenCampaign()"
       >
         <md-icon>assignment_turned_in</md-icon>
--- a/client/src/components/campaign/campaign.html
+++ b/client/src/components/campaign/campaign.html
@@ -35,7 +35,7 @@
       </md-button>
       <md-button
         aria-label="Close Campaign"
-        ng-if="!$ctrl.edit"
+        ng-if="!$ctrl.edit && !$ctrl.isCampaignClosed"
         ng-click="$ctrl.closeCampaign()"
       >
         <md-icon>assignment_turned_in</md-icon>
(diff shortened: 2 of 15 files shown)


atsirlin:
> Could you please clarify what could be wrong with that approach?
Nothing is wrong in my opinion. The existing version does not allow to close the campaign unless it reached the ranking round, and this round has been finalized. However, many of the campaigns never reach the ranking round, either because organizers chose their own way and did the final ranking without Montage, or because the campaign was one of several campaigns created for a given contest (so the ranking round for this contest took place in a different campaign, if at all). Therefore, it is absolutely crucial to be able to close campaigns at any point of time. I have about 50 open campaigns on my list that I dream to close but I can't.

mahmoud:
Sure. First off, thanks for this, I appreciate the initiative. This PR is a good starting point for discussion. Here are some thoughts:

It definitely makes sense to close a campaign if it's not proceeding any further. I'm just not sure if that should be called finalization. Finalization entails that there are final results stored somewhere in the system, and you can see that `build_campaign_report`, for instance, relies on the final round being a ranking round. This approach breaks that invariant.

Could a separate status make sense? Going from 3 to 4 statuses will mean revisiting invariants and transitions to make sure we handle the new status.

Personally, I'd definitely love this; Stephen and I see probably more campaigns than anyone else in the system, and if you thought the UI was slow before, phew! :)

atsirlin:
@mahmoud: from the user's perspective it is just the 'closed campaign', no matter if it was really finalized or simply served its purpose. It is actually a wrong assumption that each grading should end with the ranking round. At least one third of the countries don't do that and choose their best photos based on the grades, or do the ranking without Montage. I would suggest to keep it simple for now: each campaign can be active or closed, and each admin can close or re-open the campaign.

MichaelNMaggs:
See also related issues #122 #144 and #157

The technical way in which previous campaign rounds have been finalised/finished/closed/dropped isn't normally of much interest to jurors. What concerns them is seeing huge numbers of old campaigns and rounds (tens in my case) that can't be hidden, take ages to load, and make it difficult to find the single active round that you're looking for. That's especially so since the active round always appears right at the end, and requires scrolling down to the bottom through all these 'dead' campaigns. If old things of no further interest could be hidden by the juror, that would solve it.

mahmoud:
@atsirlin @MichaelNMaggs My reply was almost entirely technical in nature, relating the piggyback-on-finalize approach @baturin has proposed to hiding the campaigns. To be clear, I definitely understand the use case and want the feature, I just don't want more error cropping up down the line :)

slaporte:
I agree, I like this feature! The need to hide old campaigns is only getting greater every year that Montage is in use.

> Could a separate status make sense? Going from 3 to 4 statuses will mean revisiting invariants and transitions to make sure we handle the new status.

Yes, I'd like to use the four statuses we have for rounds: active, finalized, paused (not active or finalized), and cancelled (basically, deleted). We don't do much with campaign status right now on the backend, so I can look into this.

@MichaelNMaggs raises an important point about improving load times. We should probably separate out the info for non-active campaigns from the `/admin` and `/juror` endpoint. It's an improvement to collapse non-active stuff, but it would be even better if users don't have to wait for it either.

baturin:
Thanks for the discussion and your ideas! I have a couple of questions to clarify.
1. Are we going to put campaigns which are not actually active (as they served their purpose) and which do no conform to the rule "last round is ranking and it is finalized" to "paused" status? Or we should set them to "finalized" status (as they are actually finished)?
2. How can I cancel (set status to "cancelled") a campaign in UI? I don't see any controls responsible for that, may be I'm looking at the wrong place, could you describe what I should do? And then - why can't we just cancel all these campaigns?

mahmoud:
Great questions. So I checked and sure enough, the UI must be lacking because production only has campaigns in `active` and `finalized` status. There is logic to cancel, but it must not have been used, and that may be for the better, because that logic also cancels rounds in that campaign, and thus votes underneath that. Other parts of the system ignore canceled votes, so I don't think we want that.

It doesn't have to work that way, of course, that's just how it would work without any other changes. One approach commonly taken in these positions is "active in the last 90 days". That might solve our problem, but if the converse, "paused and old", means "hide by default", how do users see those again? Maybe an archived flag?

slaporte:
Hmm. Is archived something we'd want to flag or just a display setting? We could conceivably check the last `modified_date` for any votes in a campaign, and then default to having then have them collapsed or less prominent after they are inactive for 90 days.

I am wondering if we even need a paused state for a campaign. I think we should decide if finalized means it must have results, or merely that it's done. One basic approach would be active, finalized, and cancelled, and then use flags to note which campaigns are finalized-with-report.

client/src/components/campaign/campaign.html
```diff
@@ -35,7 +35,7 @@ <h2 layout="row" layout-align="center center" ng-if="!$ctrl.campaign.id">
       </md-button>
       <md-button
```
slaporte, line 36: I think we should also have a control option for re-activating a campaign.

client/src/components/dashboard/dashboard.js
```diff
@@ -45,11 +45,13 @@ function controller(
       .getAdmin()
       .then(data => {
         vm.isAdmin = data.data.length;
+        const isActiveCampaign = campaign =>
```
slaporte, line 48: I think this should be based on `campaign.status === 'active'`

montage/admin_endpoints.py
```diff
@@ -606,13 +606,9 @@ def finalize_campaign(user_dao, campaign_id):
...
     last_rnd = coord_dao.campaign.active_round
 
-    if not last_rnd:
-        raise InvalidAction('no active rounds')
-
-    if last_rnd.vote_method != 'ranking':
-        raise InvalidAction('only ranking rounds can be finalized')
-
-    campaign_summary = coord_dao.finalize_ranking_round(last_rnd.id)
+    campaign_summary = None
```
slaporte, line 609: Should there be a different status for a campaign that's done but doesn't have a summary?

client/src/components/campaign/campaign.component.js
```diff
@@ -34,6 +34,8 @@ function controller($filter, $q, $state, $stateParams, adminService, alertServic
         if (!vm.campaign.rounds.length) {
           vm.campaign.rounds.push({});
         }
+
+        vm.isCampaignClosed = vm.campaign.status === 'finalized';
```
slaporte, line 38: 👍

baturin:
@slaporte @mahmoud thanks for your feedback.

Personally I think we should have 3 campaign statuses: 'active' and 'finalized', and optionally 'cancelled', where:
- 'active' means our users are going to work with that campaign (create rounds, vote, view vote statistics, and so on)
- 'finalized' means that campaign has served its purpose for users, regardless of results in database and regardless of the last round type
- 'cancelled' means that campaign is garbage, e.g. created by mistake

That solution has the following advantages:
1) It directly conforms to the actual business logic, described by @atsirlin 
2) It would give us a simple fast way to split campaigns we need from those which we don't need by simple SQL query, with no joins to rounds and other tables and no complex conditions. That would help us to resolve load time problem.

As a disadvantage, we need to find and fix all places like "build_campaign_report" mentioned by @mahmoud . That is mostly a quality risk, bugs may arise.

As an additional measure, I suggest to sort all campaigns by creation date (or last modification date?) in UI - that is already in PR.

baturin:
After brief review I have not managed to find any way to view final report (which may depend on the last round results) from GUI, I see only API endpoints which are not used, and entrypoints like /admin/campaign/<campaign_id:int>/report, which are not referenced (still could be used by direct link). May be I'm checking at wrong places, could you please assist me?

baturin:
@mahmoud @slaporte do you have any comments or updates on that PR? Thanks in advance!

mahmoud:
Hey @baturin, wanted to revive this one over the holiday weekend. Noticed a few things

1. By creating two new dashboard sections for archived campaigns, the component appears to be loading the same endpoints (and the same data) twice (based on the Network tab). In some cases this might be OK, but here, Toolforge is resource-strapped, so it'd be great if we could fetch just once, and filter out the archived data on the server unless explicitly requested by client. A "View Archive" approach as done with the Juror view could work better, too.
2. There don't appear to be any backend tests, could we add a few of those to the pytest suite?

atsirlin:
@mahmoud: just a short comment. This pull request has been around for several months, but we can't discuss it forever, especially at such a slow pace. "Two dashboard sections" are not new, I have them on my screen since more than a year, and I did not see any major issue with the way they are organized or with their loading time. Let's focus on the main problem - that we need a solution to force close the campaigns. It may be a dirty solution for now, but it's better than having no solution for years.

baturin:
Hi @mahmoud!

1. As for two requests - the main dashboard already uses the same approach, sending two requests (one for admin, one for juror). I've simply reused that approach to keep code consistent, and keep API simple to use and test by splitting separate data over different requests.
2. Sure, we could add some tests once we agree how the code should work :)

mahmoud:
@baturin Got it. One request for admin and one for juror is understandable. I was seeing two for admin and two for juror.

Here's my thought, I appreciate @atsirlin's urgency, so I started a branch last night which shows Campaigns for the last 90 days and then instead of archive shows "All" (I think current/all vs. current/archived is a better pattern). It'll clean up the screen and get a faster load time. Coordinators who want to signal a campaign is cancelled or superseded can modify the name and it'll clear out of the screen over time. Because we've got some concerns over state and flags, I figured this would be the most expedient route.

Thoughts?

atsirlin:
@mahmoud: I don't understand how it is supposed to work. Could you give an example?

mahmoud:
@atsirlin so the way my local branch works is that campaigns with close dates 90 days in the past are not shown by default. You will have to click on "All Campaigns" to get to the full list. Not perfect; I'd still like to add the Archive flag and then change the default view to be "active in last 90 days + not archived" (and All continues to show all).

mahmoud:
Stephen and I had a chat about this and we're going to make a run of building a proper "archive" (extending+testing the work in this PR). Will keep you posted.

mahmoud:
Merging, continuing in #178 (Thanks @baturin and @atsirlin, see you there!)

PR #171, opened 2020-06-28
Allow any host for dev environment

baturin:
Allow any hostname, not only localhost, for development environment


mahmoud:
More local testing improvements, yay!

mahmoud:
Wish yarl were here to check, but I think this one's safe :)

PR #172, opened 2020-06-28
Specify source in config

baturin:
Add ability to use fair Commons database in development environment by config file option (BTW I have not managed to use "remote" source last time I tried).

### DIFF
--- a/montage/rdb.py
+++ b/montage/rdb.py
@@ -1526,11 +1526,7 @@ class CoordinatorDAO(UserDAO):
 
     def add_entries_from_cat(self, round_id, cat_name):
         rnd = self.user_dao.get_round(round_id)
-        if ENV_NAME == 'dev':
-            source = 'remote'
-        else:
-            source = 'local'
-        entries = loaders.load_category(cat_name, source=source)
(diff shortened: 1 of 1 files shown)


montage/rdb.py
```diff
@@ -3097,6 +3085,23 @@ def choose_eligible(eligible_users):
             'task_count_mean': mean(vote_count_map.values())}
 
 
+def get_source():
+    from utils import load_env_config
+
+    config = load_env_config()
+
+    config_source = config.get('source')
```
mahmoud, line 3093: `source` is a little vague here, can we call it `commons_source`?

An aside: `remote` and `local` are pretty bad names, too, I'll take the full blame. The proper meaning, maybe worth a comment: remote = toolforge util service (ideal name: `tools_api`), local = toolforge commons db replica (ideal name: `db_replica`).

mahmoud (approved):
Looks good, should help with local testing. Note that the Toolforge utility API used for testing is down sometimes, I usually log in and give it a restart and it runs for a few more months. Also, sometimes the schema changes out from under it, in which case we fix the code and redeploy :)

PR #173, opened 2020-06-28
Allow to specify current user in config

baturin:
Ability to specify user in config file, to get simple way to switch user in development environment, with no need to modify "mw.py", and with no need to configure OAuth.

### DIFF
--- a/montage/mw.py
+++ b/montage/mw.py
@@ -115,18 +115,22 @@ class UserMiddleware(Middleware):
                         or '/static/' in _route.pattern
                         or isinstance(_route, NullRoute))
 
-        try:
-            userid = cookie['userid']
-        except (KeyError, TypeError):
-            if config['__env__'] == 'devtest':
-                # TODO: until github.com/pallets/werkzeug/issues/1060
-                userid = 6024474
-            else:
-                if ep_is_public:
-                    return next(user=None, user_dao=None)
-                err = 'invalid cookie userid, try logging in again.'
-                response_dict['errors'].append(err)
-                return {}
+        config_userid = config.get('userid')
+        if config_userid:
+            userid = config_userid
+        else:
+            try:
+                userid = cookie['userid']
+            except (KeyError, TypeError):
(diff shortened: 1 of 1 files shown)


montage/mw.py
```diff
@@ -115,18 +115,22 @@ def endpoint(self, next, cookie, rdb_session, _route, config,
...
-                userid = 6024474
-            else:
-                if ep_is_public:
-                    return next(user=None, user_dao=None)
-                err = 'invalid cookie userid, try logging in again.'
-                response_dict['errors'].append(err)
-                return {}
+        config_userid = config.get('userid')
+        if config_userid:
+            userid = config_userid
```
mahmoud, line 120: seems like it'd be safer to also check the env, as is done down below, no? `config['__env__'] == 'devtest'`

montage/mw.py
```diff
@@ -115,18 +115,22 @@ def endpoint(self, next, cookie, rdb_session, _route, config,
...
+        else:
+            try:
+                userid = cookie['userid']
+            except (KeyError, TypeError):
+                if config['__env__'] == 'devtest':
+                    # TODO: until github.com/pallets/werkzeug/issues/1060
+                    userid = 6024474
```
mahmoud, line 127: also, maybe this should be `config.get('userid', 6024474)` for a unified devtest experience?

baturin: I have '__env__': 'dev', not 'devtest'. How should we unify that?

mahmoud: ah, good question. devtest is for the pytest run. dev is for local dev. devtest only comes from `utils.py`. With this PR, we could move the userid into that dictionary in `utils.py`, and allow userid overriding for dev and devtest. 

I don't think we want to unify dev and devtest because a stray config.dev.yaml could change the pytest behavior. Mostly I want to limit the envs where `userid` is considered. Does that make sense?

mahmoud:
Upon further review, I think we can give this one a shot. `su_to` is still respected, and this should help with local testing.

Issue #141, opened 2018-11-09
Handle common issues on import


slaporte:
When a coordinator imports files, Montage should:

- [ ] warn about deleted files #30 
- [ ] resolve moved files #97 
- [ ] resolve full Wikimedia Commons URLs
- [ ] resolve URL encoding issues
- [ ] resolve common typos (spaces or underscores at the beginning or end of the filename)
- [ ] warn users when files are not imported #108 
- [ ] save a log of files that were not imported #108  
- [ ] provide an easy tool to figure out which files from your source list are not in the round #140 
- [ ] ensure there are no `duplicate entry` errors #146

PR #175, opened 2020-07-24
Avoid fail when file does not exist

baturin:
Let's start sorting out issues from https://github.com/hatnote/montage/issues/141. That simple PR avoids montage from failure when importing renamed/removed/incorrect image files.

### DIFF
--- a/montage/loaders.py
+++ b/montage/loaders.py
@@ -126,8 +126,10 @@ def load_name_list(file_obj, source='local'):
         edicts, warnings = get_by_filename_remote(rl)
     else:
         for filename in rl:
-            edict, warnings = get_file_info(filename)
-            edicts.append(edict)
+            file_info = get_file_info(filename)
+            if file_info is not None:
+                edict, warnings = file_info
+                edicts.append(edict)
 
     for edict in edicts:
         try:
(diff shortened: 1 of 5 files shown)


montage/loaders.py
```diff
@@ -182,18 +184,25 @@ def get_entries_from_gsheet(raw_url, source='local'):
...
         files = []
         for filename in filenames:
             file_info = get_file_info(filename)
-            files.append(file_info)
+            if file_info is not None:
+                files.append(file_info)
+            else:
+                warnings.append(
+                    'file "%s" does not exist, please check that its name is spelled correctly, '
+                    'that it has not been renamed or removed' % (filename,)
```
mahmoud, line 200: This is a good, actionable error message. However, we should keep an eye out for if there are many, many missing files. Assuming there are 0-10, I think the error display method defined in the JS above should be fine.

baturin:
@mahmoud Yes, I think we should add some tests for situations like that (there are also several situations around that should be handled in a similar way). BTW, have you checked my previous test improvements https://github.com/hatnote/montage/pull/159?
As for displaying error message, that's also pretty good question, and I think yes, we should rework error display method to make it more structured and handle many errors (as there could be many files with issues like that). I'm going to think of a way once we handle all issues on backend.

mahmoud:
Cool. So, what I'm understanding is that we can roll with this for now and will improve error display in a future PR?

As for #159, somehow I missed that! Left a comment. :)

mahmoud:
@baturin just checking in on this, are we OK to merge + address errors in the next?

baturin:
@mahmoud Yeah, let's continue improving error reporting in future PRs, and merge that one.

PR #178, opened 2020-09-13
Hiding inactive campaigns (Archive flag)

mahmoud:
Rather than coopt the status field of the Campaign table, as in #169, this PR introduces a new Archive feature, which hides the campaign from  Jurors' and Coordinators' dashboards. Campaigns in any state can be archived without affecting their status.

Current status is that the backend is tested and passing. Now, need to make the modifications to the frontend.

For rollout:
  * Need to add the column + index
  * Autoarchive campaigns older than a year to save some coordinator clicking?

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "montage",
-  "version": "19.09.23",
+  "version": "20.09.17",
   "description": "Photo evaluation tool for and by Wiki Loves competitions",
   "main": "src/app.module.js",
   "repository": "https://github.com/hatnote/montage",
--- a/client/src/app.module.js
+++ b/client/src/app.module.js
@@ -26,7 +26,7 @@ import campaignNew from './components/campaign-new/campaign-new.component';
 import campaignOld from './components/campaign-old/campaign';
 
 import dashboard from './components/dashboard/dashboard';
-import archive from './components/dashboard/archive';
+import all_campaigns from './components/dashboard/all_campaigns';
 import campaignAdminBox from './components/dashboard/campaign-admin-box/campaign-admin-box.component';
 import campaignJurorBox from './components/dashboard/campaign-juror-box/campaign-juror-box.component';
 import footer from './components/footer/footer.component';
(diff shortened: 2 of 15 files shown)


mahmoud:
The commit above adds the frontend. This is basically done except for maybe some cleanup and rollout activities.

Also, it appears the double-load I spotted in the original PR is also on production. Seems like the dashboard's onInit is being executed twice, kinda like this: https://stackoverflow.com/questions/38787795/why-is-ngoninit-called-twice

slaporte:
Nice! The frontend is all working for me.

A few small things (that I don't think should hold back this PR):

- In the dashboard view, we probably want to add a header or intro that explains it's only showing active campaigns.
- `$ctrl.canCloseCampaign` should probably check if the campaign's latest round is ranking. Otherwise, it will raise an error.
- How should coordinators view closed campaigns?
- Speaking of campaign closing. Should an archived round be closable? Should closed campaigns be archivable?
- We should probably bump the client version (19.09.23!)

slaporte:
> The commit above adds the frontend. This is basically done except for maybe some cleanup and rollout activities.
> 
> Also, it appears the double-load I spotted in the original PR is also on production. Seems like the dashboard's onInit is being executed twice, kinda like this: https://stackoverflow.com/questions/38787795/why-is-ngoninit-called-twice

I think `onInit` [for the toolbar](https://github.com/hatnote/montage/blob/5d17873d594ae206ad3dece138b341c9e596b308/client/src/components/toolbar/toolbar.component.js#L28:L29) is the culprit. Admin and juror data is not necessary to load for the toolbar, and we load it again for for the dashboard. I think we can cut those two lines without a problem.

mahmoud:
All valid points.

1. I can update the header to say "Active Campaigns"
2. I thought the same about `canCloseCampaign`, but the server-returned error message is actually pretty descriptive, so I didn't foray too late.
3. Are closed campaigns not in the "all" view? If not, I'll address that.
4. I'd say ideally all archived content is paused and encased in glass until unarchived. Closed campaigns are certainly archivable, though right now the only effect I can think of would be the button changing labels.
5. 20.09.16! :)

mahmoud:
Just confirmed that closed campaigns are on the "All" page.

mahmoud:
Screenshots with one open and one closed campaign:

Screenshot from 2020-09-16 23-54-28
Screenshot from 2020-09-16 23-55-30

mahmoud:
Alright, let's get this deployed. Merging.

mahmoud:
Steps for deploying this

1. Build frontend
1. Migrate DB (add column)
1. Upload code (frontend+backend)
1. Restart
1. Test
1. One-time archive campaigns older than X months

Do ^ for dev/beta before prod.

PR #239, opened 2024-11-01
v2 version

Jayprakash-SE:


requirements.txt
```diff
@@ -4,7 +4,7 @@
 #
 #    pip-compile --annotation-style=line requirements.in
 #
--e git+https://github.com/the-maldridge/python-vote-core.git@f0b01e7e24f80673c4c237ee9e6118e8986cf0bb#egg=python3_vote_core  # via -r requirements.in
```
Jayprakash-SE, line 7: Hi @mahmoud, this is the only change that I am afraid that it can break the legacy version. I made this change because of -e parameter, docker was not able to install this in the container. With -e, docker and toolforge deployment both is working.


Reset, you can merge PR without worry.

deployment_v2.md
```diff
@@ -0,0 +1,122 @@
...
+exit
+```
+
+##### 8. Start the backend service
+```bash
+toolforge webservice python3.9 start
+```
+
+##### 9. Testing of deployment
+* Visit /meta to see the API. Example: https://montage-beta.toolforge.org/meta/
```
mahmoud, line 70:
```suggestion
* Visit /meta to see the API. Example: https://montage-beta.toolforge.org/meta/
  * In the top section, you should see that the service was restarted in the last few seconds/minutes.
```

deployment_v2.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+## Deploying new changes
+
+If montage is already deployed then you just need following to deploy new changes.
+
+##### 1. Check the instance usage
+Login to the tool webapp. Make sure, you are maintainer on the webapp instance. Use the audit log endpoint to check that the instance isn't in active use. Example: https://montage-beta.toolforge.org/v1/logs/audit
+
+This will tell latest usage of instance by audit `create_date`. You can continue if instance is not being used.
```
mahmoud, line 83: To be fair, sometimes it's in use, but there's an important bugfix and we have to push anyways. People will usually just retry after a few minutes.

dev_v2.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+### 2. Start the Frontend Build Watcher
+```bash
+cd frontend
+npm install
+npm run watch:build
+```
+
+⚠️ **Frontend does not run in dev mode**: This will:watch for changes by using `chokidar` command and automatically copied to the backend's `static` and `template` directories. So there will not be any hot reloading. Developers must refresh the browser to see updates.
```
mahmoud, line 50:
```suggestion
⚠️ **Frontend does not run in dev mode**: `npm run watch:build` above will watch for changes by using `chokidar` command and automatically copied to the backend's `static` and `template` directories. So there will not be any hot reloading. Developers must refresh the browser to see updates.
```

dev_v2.md
```diff
@@ -0,0 +1,200 @@
...
+
+Apart from `npm watch:build`, these are other commands:
+* `npm run lint`: Lint the code
+* `npm run format`: Format the code
+
+### 3. Use the Makefile to start the backend
+* Open a new termial tab and change directory to root of repo
```
mahmoud, line 57:
```suggestion
* Open a new terminal tab and change directory to root of repo
```

dev_v2.md
```diff
@@ -0,0 +1,200 @@
...
+* Copy and edit `config.dev.yaml` based on `config.default.yaml`
+* (Optional) In `config.dev.yaml` there is a line for `dev_local_cookie_value`. To get it,
+log in to the local app in your browser, and then copy the value from the
+`clastic_cookie` in the apps' cookies. This is your login cookie.
+* (Optional) Add your username as the `superuser` in the config. (This will allow you to
+add `su_to=<any user>` to the backend, if you want to test submitting as another
+juror.)
```
mahmoud, line 64: this is fine to mention, but I believe su_to is mostly used by automated tests these days :)

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  montage:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - "5001:5000"
```
mahmoud, line 7: any particular reason why we remap this here?

deployment_v2.md
```diff
@@ -0,0 +1,122 @@
...
+cd $HOME/www/python
+git clone https://github.com/hatnote/montage.git src
+```
+
+##### 4. Make the frontend build
+```bash
+toolforge webservice node18 shell -m 2G
+cd $HOME/www/python/src/frontend
+npm install
+npm run toolforge:build
```
mahmoud, line 30: Just thinking about how often things can change/break on toolforge, I'm wondering if it makes sense to document a way to build a production version locally and upload it using simple tools (without needing a whole node env)?

mahmoud (commented):
Some initial questions, comments, and typographical fixes. Haven't looked at the vue very much, this is mostly around deployment. Also, you mentioned there may be some divergences from legacy re: configs (uwsgi?), can you highlight those here please?

requirements.txt
mahmoud: Ah, that's a good callout. But it's fine to remove this, especially in production. To be extra cautious, we could create a quick backup of the legacy venv on toolforge.

Jayprakash-SE: Okay

docker-compose.yml
Jayprakash-SE: On Macbook, AirPlay functionality runs on port 5000. This always make problem to new developer. See https://forums.developer.apple.com/forums/thread/682332

deployment_v2.md
```diff
@@ -0,0 +1,122 @@
+# Montage Deployment
+
+These are instructions for deploying Montage on Toolforge.
+
+## Deploying on Toolforge from scratch
+These instructions is only first time when setuping project on Toolforge
+
+##### 1. Get thee OAuth credentials.
```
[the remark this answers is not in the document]
Jayprakash-SE: Fixed

deployment_v2.md
```diff
@@ -0,0 +1,122 @@
...
+##### 5. Create your database
+* Get the user name of database (`cat ~/replica.my.cnf`)
+* Open up MariaDB with `sql local`
+* Create a [Toolforge user database](https://wikitech.wikimedia.org/wiki/Help:Toolforge/Database#User_databases) (`create database <user>__<db name>;`), and remember the name for the config
+
+##### 6. Set up the montage config
+* Make a copy of `config.default.yaml` for your environment
+   * You may need to update `USER_ENV_MAP` in `montage/utils.py` if you need to add a new environment
+* Add the `oauth_consumer_token` and `oauth_secret_token` 
+* Add a `cookie_secret: <your randome secret>`
```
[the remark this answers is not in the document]
Jayprakash-SE: Fixed

deployment_v2.md
Jayprakash-SE: Added to doc

deployment_v2.md
Jayprakash-SE: Added this context as well in doc

deployment_v2.md
Jayprakash-SE: I played with fabric.I think we can easily zip the bundle locally and upload it to the tool with a webservice restart as we don't need to go inside pos.

mahmoud:
> > Also, you mentioned there may be some divergences from legacy re: configs (uwsgi?), can you highlight those here please?
> 
> @mahmoud See https://github.com/hatnote/montage/tree/master/config

is the concern around the slashes? has the url structure changed? i don't think we make a lot of promises about deeplinks/bookmarks, etc. We don't have public URLs besides login/home.

Other than that, the config only affects backend, if it's still used, even.

docker-compose.yml
mahmoud: Ohhh that's a good callout. Can we comment that in the file then?

```suggestion
      - "5001:5000"  # Airplay runs on port 5000 on mac. See https://forums.developer.apple.com/forums/thread/682332
```

Jayprakash-SE:
> > > Also, you mentioned there may be some divergences from legacy re: configs (uwsgi?), can you highlight those here please?
> > 
> > 
> > @mahmoud See https://github.com/hatnote/montage/tree/master/config
> 
> is the concern around the slashes? has the url structure changed? i don't think we make a lot of promises about deeplinks/bookmarks, etc. We don't have public URLs besides login/home.
> 
> Other than that, the config only affects backend, if it's still used, even.

There is no concern. The new version does not use the uwsgi.ini files but rather uses standard approach (https://wikitech.wikimedia.org/wiki/Help:Toolforge/Web/Python#Conventions).

My question was that I don't have experience with these config files so, in case of rollback. I may struggle to redeploy the app because I did the custom config.

Issue #300, opened 2025-07-25
Add alembic for db migration


Jayprakash-SE:
Add the Alembic support for database migration

PR #301, opened 2025-07-26
Issue: Add alembic support for db migrations

AadarshM07:
## Description
- Integrated Alembic support for managing database migrations.
- Updated the Docker setup to initialize and support Alembic migrations automatically during container startup.


lgelauff:
This is an implementation of #300 — thanks for taking a look at this @AadarshM07, and sorry for the slow feedback. I'm going to close this for now because a few foundational pieces are missing before this can work: the sqlalchemy.url in alembic.ini is hardcoded to the local SQLite path rather than reading from the app's config, the initial migration is empty (it needs to capture the current schema as a baseline), and alembic is unpinned in requirements.txt which risks pulling in a version incompatible with the pinned sqlalchemy==1.2.19.

PR #313, opened 2025-07-27
Add notification page and functionality

### LINKED ISSUE #291: Allow notification of jurors with incomplete rounds
This is a feature request. It could be nice to have a system within montage to send reminders to jurors if they have incomplete rounds. This could be an echo notification or a message on a talk page or an email, if it is available. Not sure which one is the most appropriate, I can see advantages and disadvantages in all. But currently, coordinators have to check periodically if jurors are finished, and notify them individually.

cr3ativ3cod3r:
Added pages for the same in frontend.
Added a route in the backend and a helper function to get an authenticated for the user.
Used the session to generate csrf token.
Made edits in the talk page of the user with the received csrf token.
Fixes #291

### DIFF
--- a/frontend/src/components/Round/NotifyForm.vue
+++ b/frontend/src/components/Round/NotifyForm.vue
@@ -2,6 +2,22 @@
   <div class="notify-users-container">
     <div class="notify-users-form">
       <h3 class="form-title">{{ $t('montage-notify-jurors') }}</h3>
+
+      <!-- Campaign and Round Name in a Row -->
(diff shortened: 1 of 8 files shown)


kcvelaga:
@Jayprakash-SE

lgelauff:
I think we need a bit more of a deeper conversation what the desirable implementation is, before writing a PR.  I'm converting this to a draft for now, while we can continue the conversation about the best implementation at the issue.

PR #358, opened 2025-11-27
fix: add client-side validation for empty fields in the create round page

### LINKED ISSUE #357: CORS Error on Round Creation
# Description
While creating a new round on a new campaign, if you didn't fill any required part of the form, the POST call will return with 500 and ends up with a CORS error. Since no alert about not filling the input boxes are given, users will just directly get a network error without understanding where they have gone wrong.

Fix : Adding an alert box if the user accidentally forgot to fill any input would work.
## Screenshot of the issue
Image

ItsAbhinavM:
# Description

This PR adds an alert system on the create route page, since the current system goes for a failed 500 POST call ending up with a CORS error, and the  users would not have an idea about the occurred error since the current alert system calls the network-error call which doesn't make sense.

Fixes #357

### DIFF
--- a/montage/rdb.py
+++ b/montage/rdb.py
@@ -118,7 +118,7 @@ flags attribute.
 
 MAINTAINERS = [
     'MahmoudHashemi', 'Slaporte', 'Yarl', 'LilyOfTheWest',
-    'Jayprakash12345', 'Ciell', 'Effeietsanders', 'Abhinavmohandas'
+    'Jayprakash12345', 'Ciell', 'Effeietsanders'
 ]
 
 """


montage/rdb.py
```diff
@@ -117,7 +118,7 @@
 
 MAINTAINERS = [
     'MahmoudHashemi', 'Slaporte', 'Yarl', 'LilyOfTheWest',
-    'Jayprakash12345', 'Ciell', 'Effeietsanders'
+    'Jayprakash12345', 'Ciell', 'Effeietsanders', 'Abhinavmohandas'
```
mahmoud: Was this intended for local development only? Being a maintainer in production means permissions to all campaigns, so we would need to run this by the WLM committee.

mahmoud (changes requested):
Hey @ItsAbhinavM! Thanks for jumping in on this. A couple questions for you. Also, I'm working on a backend fix for the CORS issue right now, as well.

ItsAbhinavM:
Sure, this is an ss  of the alert you would get if you miss filling the juror, so this will avoid the unecessary POST which results in CORS.
image

Issue #310, opened 2025-07-27
Error in creating round 3 in campaign #928


CiellB:
The jury coordinator of [#928](https://montage.toolforge.org/#/campaign/928-wiki-loves-africa-2025-in-wugn-gombe-network) [complains ](https://commons.wikimedia.org/wiki/User_talk:Ciell#Montage_tool_help)about not being able to create round 3. (And a non-descriptive error message)

What I observe is that in round 1 not all tasks were completed, but round 2 was created nevertheless.
In round 2 all tasks are completed, but they are unable to create round 3. 
* please help the coordinator to set up round 3;
* please make sure new rounds cannot be created when there are open tasks in previous rounds;
* please make sure the error in round creation is descriptive.

lgelauff:
This seems to be indeed a duplicate of #351 . Thanks for reporting. Closing now.

PR #361, opened 2026-01-04
fix: Round 3 creation bug 

### LINKED ISSUE #351: error when creating a new round
Internal server error: <ExceptionInfo [RecursionError: maximum recursion depth exceeded] (993 frames, last=Callpoint('<listcomp>', 101, 'pygraph.classes.digraph', '/data/project/montage/www/python/venv/lib/python3.9/site-packages/pygraph/classes/digraph.py', 4, '        return [ a for a in self._edges() ]'))>

and it doesn't automatically save the quorum number after saving from "add round", i have to go to "edit round" and save again...

ItsAbhinavM:
# Description

Fixes #351 
This PR fixes the issue where when we move to round 3 after finalizing round 4, it results in an unrelated error. That reason was because of a backend problem which was solved [earlier](https://github.com/hatnote/montage/commit/c7ef2d3ec32e9829a3c7355b237803df9bfa8bcb). However that change was not reflected on the frontend which made the bug alive. 

This PR updates the frontend to handle preview failures gracefully, avoids calling `preview_results` for ranking rounds, and prevents advancing a round until valid thresholds are available, ensuring the frontend reflects the backend’s constraints.


ItsAbhinavM:
I think this PR also solves this #310 issue, as it also related to round 3 creation.

frontend/src/components/Round/RoundNew.vue
```diff
@@ -392,10 +406,15 @@ watch(thresholds, (value) => {
 })
 
 onMounted(() => {
-  if (prevRound) {
-    adminService.previewRound(prevRound.id).then((resp) => {
-      thresholds.value = resp.data.thresholds
-    })
+  if (prevRound && prevRound.vote_method !== 'ranking') {
+    adminService
+  .previewRound(prevRound.id)
```
lgelauff: Could you check the indentation here?

frontend/src/components/Round/RoundNew.vue
```diff
@@ -269,6 +269,13 @@ const submitRound = () => {
       message: $t('montage-required-fill-inputs')
     });
     return;
+  } else {
+    if (formData.value.threshold === null || formData.value.threshold === undefined) {
+      alertService.error({
+        message: $t('montage-round-threshold-required')
```
lgelauff: Could you check whether this i18n key is defined in `en.json` and the other language files? It does not appear to be present — could you confirm it follows the project i18n conventions?

frontend/src/components/Round/RoundNew.vue
```diff
@@ -319,6 +326,13 @@ const submitRound = () => {
       threshold: formData.value.threshold
     }
 
+    if (!thresholds.value) {
+      alertService.error({
+        message: $t('montage-round-cannot-advance-no-votes')
```
lgelauff: Same here — could you check whether this key is defined in `en.json`?

lgelauff (commented):
The core fixes are correct: skipping `previewRound` for ranking rounds and adding `.catch()` are exactly what's needed to resolve the issue.

The PR also adds three things that should be removed:

1. **Threshold guard in `submitRound`** — this runs for all rounds before the `roundIndex === 0` check, so it fires for Round 1 too. `threshold` is never set for Round 1 (there is no previous round to advance from), so it is always `null`. Verified locally: clicking Submit on Round 1 shows the raw key string `montage-round-threshold-required` as an error and blocks round creation.
2. **Second `!thresholds` guard** — the submit button is already disabled when `!thresholds`, is this branch still reachable?
3. **Two i18n keys** (`montage-round-threshold-required`, `montage-round-cannot-advance-no-votes`) — not defined in `en.json` or the other language files (see inline comments). Verified locally: the raw key string is displayed instead of a translated message.

Please remove those three additions and resubmit — just the `.catch()` and the ranking-round skip logic are needed.

ItsAbhinavM:
@lgelauff I have made changes according to your suggestion, can you check and suggest whether it requires any more changes. Thanks!

lgelauff:
@ItsAbhinavM Could you please make a screen recording with the reproduction of the pre-PR situation and post-PR, to compare? This helps us understand how you locally tested it. Thanks!

lgelauff:
@ItsAbhinavM could you also resolve the in-line comments when you completed them? Thanks!

ItsAbhinavM:
- Pre-PR situation

https://github.com/user-attachments/assets/4c3e3a64-9a6f-4b21-aa7d-6ff75d785671

Since GitHub requires the recordings to be less than 10MB, the texts might not be very visible, the endpoint which failed is `POST	http://localhost:5001/v1/admin/round/15/advance` and the error shown was `Bad Request: Your web client or proxy sent a request that this endpoint could not understand.`
-  Post-PR situation

https://github.com/user-attachments/assets/6ce14149-3d41-4bf7-b302-89646d96c46c

Now in this edge case situation the create round button will not be available, once it is fixed the button would be available as usual. I've removed the i18n texts added in the initial commit in the PR, since that part was an unreachable code block and would anyways never get displayed.

ItsAbhinavM:
@lgelauff here is the recordings for the bug, please review and suggest whether it requires any more changes. This is what I've came up with, would love to get some suggestions regarding this since I also feel this bug  to be a bit unusual.

Jayprakash-SE (approved):

Issue #147, opened 2019-05-18
Coordinator should see which images were chosen as favorites


atsirlin:
We need an endpoint that will tell the campaign coordinator which images were chosen as favorites. The idea is to feed this information into the csv-report produced by the "Download results" button. Presently, this csv-report contains the average grade and the grade given by each juror. We should at least supply one more column to say whether the image has been favored by someone.

The bigger idea behind it is to track selected images that were particularly liked by one of the jurors. The first 'yes/no' round may include thousands of files, which makes it easy to miss high-quality images if one of the jurors accidentally presses the No button. Therefore, our jurors were asking for this option of choosing their favorite images that can be tracked during the grading process.

lgelauff:
Sharpening scope: coordinator visibility via the results CSV — distinct from #333

This issue (#147) = coordinator-facing: surface favorites in the "Download results" export so liked images aren't lost (e.g. a high-quality image a juror accidentally "No"-ed in a large round).
#333 = juror-facing: a juror's own favorites list view. Different audience, different surface.
Current state: coordinators have no visibility into favorites today — there are zero fave references in admin_endpoints.py, and the results CSV (built by make_vote_table) contains only the average and per-juror grades.

Proposed approach (concrete):

Augment the existing results CSV (/v1/admin/round/<round_id>/results/download) rather than adding a separate endpoint.
Add favorites columns:
favorites — count of active faves for that entry, and/or
favorited_by — the usernames of jurors who favorited it (more useful than a bare yes/no, and still answers the original "whether favored by someone").
Back it with a new read-only CoordinatorDAO method (e.g. get_round_fave_map(round_id)) that returns active faves grouped by entry. No schema change / no migration.
Consider mirroring the same data into the results preview screen, not just the CSV (optional, can be a follow-up).
Acceptance criteria:

not done
The results CSV includes favorites information per image.
not done
Only active faves are counted; cancelled ones are excluded.
not done
Scoped correctly to this round/campaign's entries.
not done
No migration; read-only addition.
not done
Regression test in test_web_basic.py.
Dependency: relies on the favorites campaign-scoping fixes ([#526](https://github.com/hatnote/montage/issues/526)/#527/#541) being correct, so the counts attribute to the right campaign.

PR #364, opened 2026-01-11
Feat: Link favourite images with campaign

### LINKED ISSUE #148: Favorite images should be linked to the campaign
Presently, favorite images are only linked to the juror who chose them. However, we should be able to retrieve favorite images for a given campaign. Therefore, campaign_id has to be stored for each favorite image.

ItsAbhinavM:
# Description

Fixes #148 

This PR intends to link favorite images of a round with their respective campaigns.
Earlier favorite were associated only with juror and entry, which made it inaccessible in a campaign. `campaign_id` is now added on each new favorite images. The `fave()` and `unfave()` functions have been adjusted accordingly to prevent any cross-campaign-collissions. 
In addition active favorites can be now retrieved from campaign.


lgelauff:
Thanks for looking into this. It turns out, #148 was already fulfilled. However while looking at your PR, I realized that there were two bugs you were fixing - I documented them in #526 and #527 . Maybe it's better to deal with those specifically, without the unfinished endpoint work here. If you want to address #147 we should probably discuss a bit more what the best way of exposing would be. Lets first arrive at a clear issue description there.

PR #365, opened 2026-01-20
Feat: Add edit page warning

### LINKED ISSUE #137: Warn when leaving Edit Votes page with unsaved changes
It's possible to leave the Edit Votes page without saving changes. If a jury member does that, any changes are silently lost. 

There should be a warning if the user attempts to leave the page without saving (especially as the main voting page records each vote as it's made without a separate save step).

ItsAbhinavM:
# Description

Fixes #137 

This PR consists of frontend changes, where it adds confirmation service for route or refresh page while a vote is being edited. I've also added an extra button to go back to the previous route to `/vote/roundId` on the header since the url under the round name felt like it was not enough and user friendly.

### DIFF
--- a/frontend/src/components/Vote/VoteEdit.vue
+++ b/frontend/src/components/Vote/VoteEdit.vue
@@ -183,7 +183,7 @@
 </template>
 
 <script setup>
-import { onMounted, onUnmounted, ref, watch} from 'vue'
+import { onMounted, onUnmounted, ref, watch } from 'vue'
 import _ from 'lodash'
 import { useRouter, useRoute } from 'vue-router'
 import { useI18n } from 'vue-i18n'
@@ -216,7 +216,6 @@ import ThumbDown from 'vue-material-design-icons/ThumbDown.vue'
 import Star from 'vue-material-design-icons/Star.vue'
 import ArrowExpandAll from 'vue-material-design-icons/ArrowExpandAll.vue'
 import Heart from 'vue-material-design-icons/Heart.vue'
-import ArrowRight from 'vue-material-design-icons/ArrowRight.vue'
 import ArrowLeftThick from 'vue-material-design-icons/ArrowLeftThick.vue'
 
 // Hooks
@@ -422,11 +421,11 @@ watch(locale, (newLocale) => {
 })
 
 const beforeUnloadListner = (event) => {
-  if (edits.value.length){
+  if (edits.value.length) {
     event.preventDefault()
-    event.returnValue="";
+    event.returnValue = ''
   }
-};
+}
 
 onMounted(() => {
   getRoundDetails(voteId)
@@ -438,16 +437,16 @@ onMounted(() => {
 
   handleResize()
   window.addEventListener('resize', handleResize)
-  window.addEventListener('beforeunload',beforeUnloadListner)
+  window.addEventListener('beforeunload', beforeUnloadListner)
 })
 
 const gobackToRound = () => {
   router.push(`/vote/${round.value.link}`)
 }
 
-onBeforeRouteLeave((to,from,next)=> {
+onBeforeRouteLeave((to, from, next) => {
   if (edits.value.length) {
-    if (confirm("There are unsaved changes. Are you sure you want to leave?")) {
+    if (confirm('There are unsaved changes. Are you sure you want to leave?')) {
       next()
     } else {
       next(false)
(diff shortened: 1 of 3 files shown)


lgelauff:
Thanks for looking into this @ItsAbhinavM . Could you please make a screen recording demonstrating the reproduction of the issue locally and the fix? Could you also make sure i18n is dealt with appropriately? Thanks!

ItsAbhinavM:
@lgelauff here is the video recording of the feature, this make sure that unsaved changes are not lost when the user accidentally closes the page/refreshes or even goes to another page. This also adds a back to round button on the top right which you can use instead of the hyperlink improving the ux.

https://github.com/user-attachments/assets/8d63067b-ff39-42b5-8780-fc9fd4482edf

ItsAbhinavM:
I have also added the i18n for the confirmation text in the file now.

frontend/src/i18n/en.json
```diff
@@ -186,5 +186,6 @@
   "permission-denied-message": "You do not have the required permissions to access this page.",
   "permission-denied-home": "Go to Home",
   "montage-required-voting-deadline": "Choose a valid voting deadline",
-  "montage-required-fill-inputs": "Please fill all the boxes"
+  "montage-required-fill-inputs": "Please fill all the boxes",
+  "unsaved-changes-confirmation": "Hay cambios sin guardar. ¿Seguro que deseas salir?"
```
lgelauff: Why is there Spanish text here?

frontend/src/i18n/es.json
```diff
@@ -189,5 +189,6 @@
 	"permission-denied-message": "No tienes los permisos necesarios para acceder a esta página.",
 	"permission-denied-home": "Ir a Inicio",
 	"montage-required-voting-deadline": "Elija una fecha límite de votación válida",
-	"montage-required-fill-inputs": "Por favor, rellene todas las casillas"
+	"montage-required-fill-inputs": "Por favor, rellene todas las casillas",
+  "unsaved-changes-confirmation": "Hay cambios sin guardar. ¿Seguro que deseas salir?"
```
lgelauff: The expectation is that translations happen via Translatewiki. Only add translations that you're fluent in. 
Normal expectations: only en.json and qqq.json get the new key. All 20+ non-English locale files should be reverted to their pre-PR state.

frontend/src/components/Vote/VoteEdit.vue
```diff
@@ -84,6 +84,11 @@
       >
         <content-save-outline style="font-size: 6px" /> {{ $t('montage-round-save') }}
       </cdx-button>
+
+      <cdx-button weight="quiet" action="default" @click="gobackToRound">
+        <ArrowLeftThick />
+        <span>Back to Round</span>
```
lgelauff: Add i18n

lgelauff (changes requested):
I think the i18n is not exactly right. I have submitted some improvements on the documentation of this workflow - please see #530

frontend/src/i18n/es.json
ItsAbhinavM: Alright, now I get it.  I did the same PR in one of an old merged PR, so I thought this is the preferred way. Apologies for doing it manually. I'll fix this in a new commit. Thank you for informing.

frontend/src/components/Vote/VoteEdit.vue
ItsAbhinavM: sure!

frontend/src/i18n/en.json
ItsAbhinavM: It was a mistake, will remove it.

ItsAbhinavM:
@lgelauff thank you for the previous review regarding i18n, I did the manual method for one of my old PRs which was merged, so I thought it would be a preferred manner for adding i18n. Now I've fixed it and is only on `en.json` and `qqq.json` . Have also added i18n for `back to round` text as well.

frontend/src/i18n/es.json
lgelauff: it might be that you saw a PR before we transitioned to translatewiki. Hopefully the documentation will help! Please resolve when you've completed this.

ItsAbhinavM: its done!

Jayprakash-SE (approved):

PR #367, opened 2026-01-22
FEAT: add alembic for db migrations

### LINKED ISSUE #300: Add alembic for db migration

ItsAbhinavM:
# Description
Fixes #300 
This PR introduces alembic for DB migrations, to the backend.


lgelauff:
This is also an implementation of #300 — thanks for the contribution @ItsAbhinavM.

I'm going to close this as well. Like the recently closed #301, the core pieces are missing: sqlalchemy.url in alembic.ini is hardcoded to the local SQLite path instead of reading from the app's config, alembic is unpinned in requirements.txt (risking incompatibility with the pinned sqlalchemy==1.2.19), and the initial migration is empty rather than capturing the current schema as a baseline.

Additionally, target_metadata in env.py is left as None (the default boilerplate), meaning autogenerate would produce nothing useful — this is actually a step back from #301 which at least wired in rdb.Base.metadata.

For future PR's, it would help us a lot if you can make a screen recording of the PR improvements compared to before and in action in a local setup.

PR #376, opened 2026-02-21
Fixed backend package versionconflict, and added instructions

harshithaMMPS333:
I faced some errors while setting up the backend packages, especially with tox and pluggy

**- problem**
Project uses
tox 3.15.0 and supports pluggy < 1
but, pip installs the latest versions of the packages so the same this happened while installing pluggy, pip had installed the latest version of pluggy

**- Cause**
Not able to install packages and dependencies for backend

there were no proper instructions provided 
and was stuck there trying to figure-out what was the reason behind that huge, massive error. took the help of AI(ChatGPT) to understand the Error, and for troubleshooting, brainstorming, and code generation to resolve the version conflict.

**- Changes made**
Screenshot 2026-02-21 at 1 44 42 PM

**- Suggestions**
It would be clear if we provide instructions in README.md
(I can create a PR for that)

All the instructions and final changes is written and modified by me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ celerybeat-schedule
 
 # virtualenv
 venv/
+.venv*/
 ENV/
 
 # Logs
@@ -128,6 +129,10 @@ coverage
 *.sw?
 
 *.tsbuildinfo
+
+# Claude Code and local temp files
+.claude/
+tmp/
 
 # Spyder project settings
 .spyderproject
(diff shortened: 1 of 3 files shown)


lgelauff (changes requested):
Thanks for tracking this down @harshithaMMPS333 — the `pluggy<1` constraint seems worth adding.

The setup instructions in `requirements-dev.txt` should be dropped though — that's not the right place for them, and the venv name (`venv`) doesn't match what the project uses (`.venv311`). I would appreciate your input on #511 though. 

So the ideal version of this PR is just the one line:
```
pluggy<1
```

lgelauff (approved):

Issue #363, opened 2026-01-08
Cannot save Ranking Round with over 100 images


strainu:
I have a Ranking Round with 107 images. I cannot find a way to display all images. When trying to save 100 images, I get "Bad Request: must submit all rankings at once. (expected 107 submissions, got 100.)". Is there a way around this except changing the round type?

lgelauff:
I dont think we should do this - the signal of a ranking becomes less and less for too big groups.

ayushshukla1807:
I traced this to two bugs in `juror_endpoints.py` that combine to produce the error:

1. **`get_tasks_from_round`** was using `MAX_RATINGS_SUBMIT = 100` as the task fetch count for ranking rounds. So a round with 107 images only served 100 tasks to the frontend — 7 images were invisible and unrankable.

2. The all-or-nothing submission check then correctly expected 107 rankings from the database but only received 100 from the client, producing: `"expected 107 submissions, got 100."`

Fix in PR #495 — introduces `MAX_RANKING_TASKS = 2000` for task retrieval and moves the 100-per-batch guard so it only applies to rating/yesno rounds (ranking rounds must submit the full ballot atomically regardless of size).

PR #377, opened 2026-02-22
Fix ranking round error message

### LINKED ISSUE #373: Return informative error when trying to save ranking round with 100+ images
It is currently not permitted to include more than 100 images in a ranking round for usability and mental burden reasons. (see also #363 ). 

Currently, the error is not very informative. 

Image

Please return a meaningful error to the organizer when they try to start the round with more than the required minimum (currently set at 100 I believe, please verify, may also be 50).

Oyelakin-Mercy:
Fixes [#373](https://github.com/hatnote/montage/issues/373) and [#363](https://github.com/hatnote/montage/issues/363)

Previously, organizers who tried to activate a ranking round with more than 100 images would initially not receive any error message. However, Jurors would not be able to submit as there is a limit of 100 entries. Also, there is a generic 400 error message when organizers preview their ranking round before all ballots are submitted. 

This has now been fixed with the following changes:

- `admin_endpoints.py:` Added two validation checks to raise an error telling the organizer to adjust their threshold to be <= 100 entries, and to adjust their threshold if fewer than 2 entries advance from the rating round.

- `src/services/api.js:` Updated the Axios response interceptor in `api.js` to extract the `errors` field from the backend response and set it as the error message before rejecting the promise.

Screenshot 2026-02-22 095711

### DIFF
--- a/frontend/src/services/api.js
+++ b/frontend/src/services/api.js
@@ -48,6 +48,9 @@ const addInterceptors = (instance) => {
       const loadingStore = useLoadingStore()
       loadingStore.setLoading(false)
 
+      const message = error.response?.data?.errors
+      if (message) error.message = message
+
       return Promise.reject(error)
     }
   )
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -604,6 +604,20 @@ def advance_round(user_dao, round_id, request_dict):
     # TODO: inherit round config from previous round?
     adv_group = coord_dao.finalize_rating_round(round_id, threshold=threshold)
 
+    if len(adv_group) > 100:
+        raise InvalidAction(
+            'There are too many entries advancing (%s). Adjust your threshold to '
+            'bring the number of advancing entries to 100 or fewer.'
+            % len(adv_group)
+        )
+
+    if len(adv_group) < 2:
+        raise InvalidAction(
+            'There are not enough entries advancing (%s). Lower your threshold to '
+            'allow at least 2 entries to advance to the ranking round.'
+            % len(adv_group)
+    )
+
     next_rnd = coord_dao.create_round(**nrp)
     source = 'round(#%s)' % round_id
     params = {'round': round_id,
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -604,10 +604,10 @@ def advance_round(user_dao, round_id, request_dict):
     # TODO: inherit round config from previous round?
     adv_group = coord_dao.finalize_rating_round(round_id, threshold=threshold)
 
-    if len(adv_group) > 100:
+    if len(adv_group) > 20:
         raise InvalidAction(
             'There are too many entries advancing (%s). Adjust your threshold to '
-            'bring the number of advancing entries to 100 or fewer.'
+            'bring the number of advancing entries to 20 or fewer.'
             % len(adv_group)
         )
 


montage/admin_endpoints.py
```diff
@@ -604,6 +604,20 @@ def advance_round(user_dao, round_id, request_dict):
     # TODO: inherit round config from previous round?
     adv_group = coord_dao.finalize_rating_round(round_id, threshold=threshold)
 
+    if len(adv_group) > 100:
```
mahmoud: Ranking really stops working above 40, and ideally would be around 20 for the best ranking. Wdyt about reducing this number?

mahmoud (changes requested):

montage/admin_endpoints.py
userAdityaa: > Ranking really stops working above 40, and ideally would be around 20 for the best ranking. Wdyt about reducing this number?

Yes, that makes sense. Since the algorithm has a time complexity of up to O(n³), the current changes are valid. However, documenting the rationale behind the algorithm choice, especially why 20 is considered the optimal ranking threshold, would make the decision much clearer. At the very least, adding a comment alongside the changes would help.

lgelauff: 'really stops working' here should be read as "stops being useful". Users have limited attention span, and ranking long lists of images has a reduced value as the list gets longer.

montage/admin_endpoints.py
```diff
@@ -604,6 +604,20 @@ def advance_round(user_dao, round_id, request_dict):
     # TODO: inherit round config from previous round?
     adv_group = coord_dao.finalize_rating_round(round_id, threshold=threshold)
 
+    if len(adv_group) > 20:
+        raise InvalidAction(
+            'There are too many entries advancing (%s). Adjust your threshold to '
+            'bring the number of advancing entries to 20 or fewer.'
+            % len(adv_group)
+        )
```
lgelauff, line 612: Please don't hardcode messages into the software. Use our i18n approach. I improved its documentation in #530 .

montage/admin_endpoints.py
```diff
@@ -604,6 +604,20 @@ def advance_round(user_dao, round_id, request_dict):
     # TODO: inherit round config from previous round?
     adv_group = coord_dao.finalize_rating_round(round_id, threshold=threshold)
 
+    if len(adv_group) > 20:
```
lgelauff, line 607: This should not be a hard stop. There should probably be two messages: one hard message, which is set by a hard constraint - it is not possible to rank more than K images here. 
The other is more of a recommendation: we suggest that you limit this to 25 or fewer images.

lgelauff (changes requested):

lgelauff:
Thanks for working on this! I don't think it touched #363 but it does #373. Please look at our comments above. I'll convert to draft in the meanwhile.

Two additional things to address:

1. **`api.js` reads the wrong field.** When `InvalidAction` is raised, clastic serializes it as `{"code": 400, "detail": "...", "message": "Bad Request"}` — there is no `errors` field. So `error.response?.data?.errors` will always be `undefined` for the new validation checks, meaning the "too many entries" and "too few entries" messages will never reach the user. The correct field to read is `error.response?.data?.detail`. (The `errors` field does exist in the preview endpoint's manual return, but that is a separate code path and it is a plain string, not a list.)

2. **Validation order** — the `> 20` and `< 2` checks fire after `finalize_rating_round` has already marked the round as closed. It only recovers because the session rolls back on a 4xx response. Better to check the advancing group size before finalizing.

When you are finished, please submit a screen recording of the reproduction locally and your fix.

frontend/src/services/api.js
```diff
@@ -48,6 +48,9 @@ const addInterceptors = (instance) => {
       const loadingStore = useLoadingStore()
       loadingStore.setLoading(false)
 
+      const message = error.response?.data?.errors
+      if (message) error.message = message
```
lgelauff, line 52: Could you check the data type?

montage/admin_endpoints.py
```diff
@@ -604,6 +604,20 @@ def advance_round(user_dao, round_id, request_dict):
...
+        raise InvalidAction(
+            'There are too many entries advancing (%s). Adjust your threshold to '
+            'bring the number of advancing entries to 20 or fewer.'
+            % len(adv_group)
+        )
+
+    if len(adv_group) < 2:
+        raise InvalidAction(
```
lgelauff, line 615: Lets keep the indentation consistent

lgelauff (changes requested):

lgelauff:
Thanks for working on this! The underlying bugs are now covered by other PRs — [#495](https://github.com/hatnote/montage/pull/495) fixes the ranking round 100-image submission limit ([#363](https://github.com/hatnote/montage/issues/363)) and [#548](https://github.com/hatnote/montage/pull/548) fixes the generic 400 on ranking round stats ([#373](https://github.com/hatnote/montage/issues/373)). Closing in favor of those.

Issue #372, opened 2026-02-08
Skip function not returning photos for later voting


dinosaur84:
Hello! 

I am a coordinator and having some issues because any images that have been skipped are not showing up later for viewing. Myself and another of my jurors have had this problem so we haven't been able to finish the round annoyingly. Any tips?

<img width="1785" height="297" alt="Image" src="https://github.com/user-attachments/assets/5d4eec95-d329-4f0f-be2c-64f6401ffb2f" />

<img width="1486" height="187" alt="Image" src="https://github.com/user-attachments/assets/06701742-800b-4d53-b122-a83e37c68354" />

lgelauff:
closing as duplicate of #371

Issue #371, opened 2026-02-03
"Vote Later" doesn't let you vote later


zacharyb02:
For Wiki Loves Clermont-Ferrand campaing, many jurors used the "vote later" for a Yes/No round. the image doesn't show after and it says that it's all done . See the screenshot below

<img width="721" height="89" alt="Image" src="https://github.com/user-attachments/assets/bb970024-0438-4348-87c0-cf8d50b78fa3" />

Even when the juror tried to edit the previous votes, the one that was labeled as later doesnt show up. But the problem is that the progress is not 100% for the admin compaing . see the screenshot below

<img width="163" height="96" alt="Image" src="https://github.com/user-attachments/assets/5756d387-950b-4fd1-9327-41a1da9e42b1" />

tejaswini-jawale:
Hi! I’m interested in working on this issue as part of my contributions toward GSoC 2026.

I plan to start by reproducing the problem locally to understand the current behavior of the “Vote Later” functionality and why it does not allow users to actually defer voting as expected. After identifying the root cause, I’ll review the relevant logic handling the voting flow and state management to ensure that selecting “Vote Later” properly preserves the item and allows the user to return and vote at a later stage.

My approach will be:

Reproduce the issue and trace the code path responsible for the “Vote Later” action.

Identify where the intended deferred voting state is not being preserved.

Implement a fix so the item remains available for later voting instead of being skipped or discarded.

Add/extend tests (if applicable) to prevent regressions.

Verify the behavior in both normal and edge cases.

Please let me know if there are any specific guidelines or parts of the codebase I should review before starting.

Thanks!

lgelauff:
Steps to reproduce  based on my understanding
                                                   
1. Log in to Montage as a juror in an active Yes/No (or rating) round         
2. On the first image, click "Vote Later" (skip)
3. Continue voting on the remaining images until no more images appear and the round shows as complete                                            
4. Observe: the skipped image never reappears — it is permanently lost from your queue, even though you never voted on it                       
5. From the coordinator view, the round shows as incomplete (the skipped taskis unvoted), but the juror has no way to access it                            
   
## Expected behaviour                                                         
Once all other images are voted on, the skipped image should reappear so the juror can complete the round.

## Root cause                                                       
[...]

PR #378, opened 2026-02-22
Fix vote later permanently hiding skipped images in voting rounds

Oyelakin-Mercy:
Fixes [#371](https://github.com/hatnote/montage/issues/371) and [#372 ](https://github.com/hatnote/montage/issues/372)

This PR fixes the "Vote Later" bug that hid images permanently from being attended to later in the yes/no and rating rounds. It showed "all done" on the Juror's end when it was not actually complete, and less than 100% on the organizer's end, which led to not being able to finalize the round. 

This has now been fixed with the following changes:

* `montage/rdb.py:` I updated `skip_voting` to store the skipped vote IDs as a list instead of a single value, so multiple images can be skipped without replacing each other.
* `montage/rdb.py:` I updated the `get_tasks_from_round` to show the non-skipped tasks first, and only return skipped tasks when all other tasks are complete, so the skipped images are always re-queued for the juror to vote on later.
* `montage/rdb.py:` I fixed the `skip` property on `RoundJuror` to read from the new `flags['skipped']` key instead of the old `flags['skip']` key.

## Note on data migration
Any jurors who used "Vote Later" before this fix have `flags = {'skip': <vote_id>}` in the database (old format). The new code ignores that key, so their previously skipped images will show again, which is the correct behavior. There will be no data loss and no error risk.

Before:
Screenshot 2026-02-22 131207
Screenshot 2026-02-22 131142

After:
Screenshot 2026-02-22 132705

### DIFF
--- a/montage/rdb.py
+++ b/montage/rdb.py
@@ -486,8 +486,9 @@ class RoundJuror(Base):
 
     @property
     def skip(self):
-        skip = self.flags.get('skip')
-        return skip
+        if not self.flags:
+            return []
+        return self.flags.get('skipped', [])
 
     def get_count_map(self):
         rdb_session = inspect(self).session
--- a/montage/tests/test_web_basic.py
+++ b/montage/tests/test_web_basic.py
@@ -957,3 +957,89 @@ def submit_ratings(client, round_id, coord_user='Yarl'):
     # get all the jurors that have open tasks in a round
     # get juror's tasks
     # submit random valid votes until there are no more tasks
+
+def test_vote_later_reappears(api_client):
+    """
+    Regression test for #371 / #372: skipped tasks should reappear after
+    the juror exhausts all remaining non-skipped tasks.
+    """
+    fetch = api_client.fetch
+
+    fetch('maintainer: add organizer',
+          '/admin/add_organizer',
+          {'username': 'Yarl'})
+
+    resp = fetch('get default series', '/series')
+    series_id = resp['data'][0]['id']
+
+    resp = fetch('organizer: create campaign',
+                 '/admin/add_campaign',
+                 {'name': 'Vote Later Repro',
+                  'coordinators': [u'LilyOfTheWest',
+                                   u'Slaporte',
+                                   u'Yarl'],
+                  'open_date': '2015-09-01 17:00:00',
+                  'close_date': '2015-10-01 17:00:00',
+                  'url': 'http://hatnote.com',
+                  'series_id': series_id},
+                 as_user='Yarl')
+
+    resp = fetch('coordinator: get admin view', '/admin', as_user='LilyOfTheWest')
+    campaign_id = resp['data'][-1]['id']
+
+    resp = fetch('coordinator: add yesno round',
+                 '/admin/campaign/%s/add_round' % campaign_id,
+                 {'name': 'Vote Later Test Round',
+                  'vote_method': 'yesno',
+                  'quorum': 1,
+                  'deadline_date': '2025-10-20T00:00:00',
+                  'jurors': [u'Slaporte']},
+                 as_user='LilyOfTheWest')
+    round_id = resp['data']['id']
+
+    fetch('coordinator: import entries',
+          '/admin/round/%s/import' % round_id,
+          {'import_method': 'category',
+           'category': 'Images_from_Wiki_Loves_Monuments_2015_in_Albania'},
+          as_user='LilyOfTheWest')
+
+    fetch('coordinator: activate round',
+          '/admin/round/%s/activate' % round_id,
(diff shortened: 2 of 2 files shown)


lgelauff:
I wrote a regression test for #371/#372 (in collaboration with Claude) and wanted to share it here. The PR passes it, but note this does not qualify as a full review.

**Without this PR**: the test fails with:
```
AssertionError: BUG #371: skipped task (vote_id=1) never reappeared. Tasks returned: []
```

**With this PR applied**: the test passes.

The test can be added to `montage/tests/test_web_basic.py`:

```python
def test_vote_later_reappears(api_client, mock_external_apis):
    """
    Regression test for #371 / #372: skipped tasks should reappear after
    the juror exhausts all remaining non-skipped tasks.
    """
    fetch = api_client.fetch

    resp = fetch('get default series', '/series')
    series_id = resp['data'][0]['id']

    resp = fetch('organizer: create campaign',
                 '/admin/add_campaign',
                 {'name': 'Vote Later Repro',
                  'coordinators': [u'LilyOfTheWest',
                                   u'Slaporte',
                                   u'Yarl'],
                  'open_date': '2015-09-01 17:00:00',
                  'close_date': '2015-10-01 17:00:00',
                  'url': 'http://hatnote.com',
                  'series_id': series_id},
                 as_user='Yarl')

    resp = fetch('coordinator: get admin view', '/admin', as_user='LilyOfTheWest')
    campaign_id = resp['data'][-1]['id']

    resp = fetch('coordinator: add yesno round',
                 '/admin/campaign/%s/add_round' % campaign_id,
                 {'name': 'Vote Later Test Round',
                  'vote_method': 'yesno',
                  'quorum': 1,
                  'deadline_date': '2025-10-20T00:00:00',
                  'jurors': [u'TestJurorVoteLater']},
                 as_user='LilyOfTheWest')
    round_id = resp['data']['id']

    fetch('coordinator: import entries',
          '/admin/round/%s/import' % round_id,
          {'import_method': 'category',
           'category': 'Images_from_Wiki_Loves_Monuments_2015_in_Albania'},
          as_user='LilyOfTheWest')

    fetch('coordinator: activate round',
          '/admin/round/%s/activate' % round_id,
          {'post': True}, as_user='LilyOfTheWest')

    resp = fetch('juror: get tasks', '/juror/round/%s/tasks' % round_id,
                 as_user='TestJurorVoteLater')
    all_tasks = resp['data']['tasks']
    assert len(all_tasks) >= 2, 'need at least 2 tasks to test skip'

    skip_vote_id = all_tasks[0]['id']
    fetch('juror: skip first task',
          '/juror/round/%s/tasks/skip' % round_id,
          {'vote_id': skip_vote_id},
          as_user='TestJurorVoteLater')

    for _ in range(100):
        resp = fetch('juror: get remaining tasks',
                     '/juror/round/%s/tasks' % round_id,
                     as_user='TestJurorVoteLater')
        remaining = resp['data']['tasks']
        if not remaining:
            break
        if any(t['id'] == skip_vote_id for t in remaining):
            break
        for task in remaining:
            fetch('juror: vote on task',
                  '/juror/round/%s/tasks/submit' % round_id,
                  {'ratings': [{'vote_id': task['id'], 'value': 1}]},
                  as_user='TestJurorVoteLater')

    resp = fetch('juror: get tasks after voting all others',
                 '/juror/round/%s/tasks' % round_id,
                 as_user='TestJurorVoteLater')
    final_task_ids = [t['id'] for t in resp['data']['tasks']]

    assert skip_vote_id in final_task_ids, (
        'BUG #371: skipped task (vote_id=%s) never reappeared. '
        'Tasks returned: %r' % (skip_vote_id, final_task_ids)
    )
```

Would be good to include this in the PR to guard against regressions.

Oyelakin-Mercy:
Hi @lgelauff, thank you so much for your review. I have tried running the test as written, but I think we are missing the `mock_external_apis` setup in the repo. pytest could not find it. Also, the `TestJurorVoteLater` does not exist as a user in a new test database.

To get it to run, I removed the `mock_external_apis` from this function `def test_vote_later_reappears(api_client, mock_external_apis):`, added a step to add an organizer `add organizer` for Yarl since the test database is empty when starting, and Yarl needs organizer permissions to create a campaign. Then I replaced the `TestJurorVoteLater` with an existing user, `Slaporte`, since unknown users would cause an error. The test now works. 

Could you share the `mock_external_apis` setup? It seems to be using a dedicated test user and mocking external API calls instead of making real network requests to Wikimedia itself. With it, we could use your original test exactly as written.

I have fixed the `skip` property on `RoundJuror` to return `flags.get('skipped', [])` instead of `flags.get('skip')`, added the data migration note to the PR description, and added the missing newline at the end of `rdb.py`. Thank you!

PR #380, opened 2026-02-25
fix: Round data invalid reloader on round deletion

### LINKED ISSUE #379: Invalid router.reload() in round deletion
# Description
The current `deleteRound()` function in `RoundEdits` uses `router.reload()` every time when a round is getting deleted, but `useRouter` never had a reload function within it, this results in error and we no change will be displayed until we manually refresh the page.

ItsAbhinavM:
# Description
Fixes #379 
Vue js uses SPA routing for navigation, and currently we were using `router.reload()` which does not exists. This PR fixes it with a relevant data reload instead of web page reload.


lgelauff:
Basically a duplicate with #435

PR #381, opened 2026-02-26
Fix: Round image import failing on creation

ItsAbhinavM:
# Description
This PR fixes the major issue of round creation when all the parameters are not met. This creates a chain of problems related to round creation. Even though the backend was perfect.

The endpoint for submitting a create round request is `/admin/campaign/<campaign_id:int>/add_round` and according to the backend function `_prepare_round_params()` the main parmeters required for a round creation are these : `['jurors', 'name', 'vote_method', 'deadline_date']`. Apart from that I also felt that we could add category in this required parameter group as I felt that it will not make any sense if there is category to import from. The main changes just includes a validation to prevent form submission when any requried parameter is left empty, on top of it the `add round` button is also deactivated.

The changes required were only in frontend. I've tested on multiple patterns in this and resulted in zero bug reproduction.


ItsAbhinavM:
Hey @mahmoud @Jayprakash-SE, could you review this PR.

This is a very small bug that has been present in Montage for quite some time. Since none of the PRs addressing it have been merged yet, people keep submitting new PRs for the same fix, which is starting to create a pile of duplicates.

If this PR looks good, it would be great to merge it and close the other related PRs. We really only need one fix for this issue, once it’s merged, the remaining PRs can simply be closed and it should prevent further duplicates from being submitted.

Thanks!

lgelauff:
Thanks for the effort here. A video reproducing the current master and your improvement would be helpful. A few specific concerns I see:

- This seems a duplicate with #471 . They seem to take diverging approaches. However, it seems more logical to follow the established pattern in `NewCampaign.vue` and `ViewCampaign.vue` — both use Zod
- i18n as discussed in other PRs. 

will do a more close review if we decide to continue with this.

Converting to draft.

ItsAbhinavM:
@lgelauff , since the additions are simple I think this snippet would cover the confusions. This is how the parameters would be alerted if you didn't fill any of them and the round will not be allowed to created until all the parameters are filled.
image
If you think this is enough, I'll add one more commit to add the i18n for it. I'll add it once the PR is changed from draft or ready to go from your side. Thanks! :)

lgelauff:
@ItsAbhinavM I believe with the integrated changes, this is now limited to a frontend improvement, right? Please check if there's a simple path to implementing the feedback. The double error message mechanism (toast & in-line messages) is a bit confusing though, but the red background is nice. Maybe @Jayprakash-SE has stronger opinions.

PR #382, opened 2026-02-27
feat: add front and back arrow to navigate through images in yes/no and rating rounds

### LINKED ISSUE #220: Support left arrow key in vote mode to return to the last image
In the voting pane, the instructions are quite clear and helpful:

image

...but I found myself more than once wishing to return to a photo I had just rated, to mull over the decision a bit further.

Oyelakin-Mercy:
Fixes [#220](https://github.com/hatnote/montage/issues/220)

**What changed:**
- Added previous/next navigation arrows overlaid on the image in yes/no and ranking rounds
- Next arrow only appears after the current image has been voted on
- Previous arrow appears from the second image onwards, which allows free backwards navigation
- The Vote state (Accept/Decline) is remembered and visually highlighted when going back to a previously voted image

**How it works:**
- `votedIds` Map tracks voted image IDs and their vote value
- `currentVote` computed shows the stored vote for the currently displayed image
- `goPrev` / `goNext` navigate through the local image batch without triggering votes or skips
- Vote buttons use `:weight="normal"` when the image has been previously voted on, which makes the selection visually clear
Screenshot 2026-02-26 235802
Screenshot 2026-02-26 235822
Screenshot 2026-02-26 235848

### DIFF
--- a/frontend/src/components/Vote/VoteRating.vue
+++ b/frontend/src/components/Vote/VoteRating.vue
@@ -16,6 +16,22 @@
         @error="handleImageLoad"
       />
       <cdx-button
+        weight="quiet"
+        class="vote-nav-prev"
+        @click="goPrev"
+        v-if="rating.currentIndex > 0"
+      >
+        <arrow-left-thick class="icon-small" />
+      </cdx-button>
+      <cdx-button
+        weight="quiet"
+        class="vote-nav-next"
+        @click="goNext"
+        v-if="votedIds.has(rating.current.id) && rating.currentIndex < images.length - 1"
+      >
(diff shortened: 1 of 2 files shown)


lgelauff:
Could you help me understand the logic of having two right-arrows? I already see one on the top-right. And there's a third 'skip'. How can we avoid that getting confusing?

lgelauff:
Also, please show a screen recording with both the current master and your improvement in action. Make sure to test what happens if people repeatedly move back and forth and vote again. 

This likely also directly conflicts with some other current PRs such as #492  and we'll have to consider that carefully.

PR #386, opened 2026-02-27
feat: add image resolution controller in create new round

### LINKED ISSUE #289: Campaign settings in round creation do not match setting as displayed after the round creation
Some settings are not in the selection menu when creating the rounds. 
Are they pre-set somewhere? Best is to have them the same!

Image

Image

### LINKED ISSUE #323: Resolution threshold can not be set
The threshold value for the image resolution can not be set during the import. The default value of 2 MP is used implicitly.

ItsAbhinavM:
# Description
Fixes #289 and Fixes #323 
According to these issues the major concern was the inability to update certain parameters on round creation notably, `min_resolution` and `file_type`. I've checked both frontend and backend files for this and came with this conclusion.

The backend is perfectly fine as `DEFAULT_ROUND_CONFIG` is being updated on `create_round` function and it includes all the parameters including `min_resolution`. So the backend is fine. The only change we need to add was in `RoundNew` file and that's what I did.

This is how new round looks like:
image

image

Database log for `min_resolution`
image

### DIFF
--- a/frontend/src/components/Round/RoundNew.vue
+++ b/frontend/src/components/Round/RoundNew.vue
@@ -115,8 +115,16 @@
                 </cdx-checkbox>
               </cdx-field>
               <cdx-field v-if="formData.config.dq_by_resolution">
-                <cdx-text-input v-model="formData.config.min_resolution" input-type="number" />
+                <cdx-text-input 
+                  v-model="formData.config.min_resolution" 
+                  input-type="number"
+                  :min="100000"
+                  :step="100000"
+                  placeholder="2000000" />
                 <template #label>{{ $t('montage-round-min-resolution') }}</template>
+                <template #help-text>
+                  <p>{{ $t('montage-round-min-resolution-help') }}</p>
+                </template>
               </cdx-field>
             </div>
           </div>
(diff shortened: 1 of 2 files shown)


lgelauff:
Could you show this in action in a video? I'm not exactly sure what the process is like here. Could you also suggest a clean test and what the expected outcome would be?

ItsAbhinavM:
@lgelauff , here is the recording of the feature. It is basically an input box to update the min resolution param for round creation. Currently all the rounds being made is just the default value, and juror's  or admins cannot customize it.

https://github.com/user-attachments/assets/07702644-358f-4d06-b9cc-4ee8fd6040a6

lgelauff:
Thanks! Did you test if the threshold is being applied as expected?

ItsAbhinavM:
> Thanks! Did you test if the threshold is being applied as expected?

Yes, I even checked it in the db and it is reflected perfectly. I have also attached a screenshot of the db as proof on the [description](https://github.com/hatnote/montage/pull/386#issue-4001731252) where I tried setting the `min_resolution` as just 100.

lgelauff:
ok, please try deploying to beta to make sure it works there too.

PR #393, opened 2026-02-28
fix: keep form open and add retry option when import returns no entries

### LINKED ISSUE #208: Montage File List URL image load interface is broken
When loading images into a new campaign using an URL the load finishes with a (fatal) error message:
**Import Warning/ import issues - no entries imported - all entries disqualified by round settings**

No images are loaded, the campaign stays empty...

See https://phabricator.wikimedia.org/T279386

Currently we can only load the image list using 1 single category, or via copy/paste of a verbatim list of file names.

Oyelakin-Mercy:
This PR fixes the UX issue where the selected category entries are all disqualified, and the round gets created with no entries. This leaves an empty round behind with no way to recover from the same form. For example, I tried creating a yes/no round with this category, "Images from Wiki Loves Monuments in Belgium", which only has subcategories and no images.
This also fixes #208 

**Problem:**
1. Round is created in the database 
2. Import runs but returns 0 entries
3. The warning dialog is shown
4. User clicks OK
5. The form closes, and an empty round is left behind

**Changes:**

`frontend/src/components/Round/RoundNew.vue`: 
- **Keep form open on failed import**: When there is an import warning, the dialog now has a "Fix Import" button that keeps the form open instead of closing it
- **Retry import on same round**: Added `createdRoundId` ref to track an already-created round. On retry, the existing round is reused instead of creating a duplicate empty round
- **Delete on cancel**: If the user cancels after a failed import, `cancelRound()` is called automatically to delete the already created empty round before closing the form.

`frontend/src/i18n/en.json`:
- Added `montage-round-retry-import` translation key for the "Retry Import" button label shown after a failed import

`montage/admin_endpoints.py`:
- Improved warning messages to be clearer:
  - Empty import: `'No entries were imported. Please fix your import source and try again, or click "Cancel" to delete the empty round.'`

**To test:**

- Create a round with a category that exists but has no images (the form stays open with a clear warning shown).
- Fix the category and click "Retry Import". This imports the retries on the same round, and no duplicate round is created. 
- Or click cancel after a failed import so that the empty created round will be deleted automatically
- Create a round with a category that has images. This will work as before.

image


lgelauff:
It seems that there's a few concerns being mixed here: 
* keeping the form open: 
* error message
* file_names split

I would propose to discuss the keeping the form open separately - it is not obvious to me that this would be an improvement. For example, this could give concerns with very long loading times. We should carefully think through that. 

From my understanding, the split of file_names might cause trouble on retry unless you guard for that. Could you test for that carefully? 

The error message is a good quick improvement though. I would focus first on that. The error message could be more informative (please double check!), because it seems to me that it's effectively triggered by the fact that we expect the url to be a gist url or google sheet. Maybe our instructions and error message should make this more explicit. It would be even better to add a front end check for this that immediately informs the user as they enter the url.

lgelauff:
(converting to draft for now)

lgelauff:
Closing for now. Feel free to open a new PR keeping the above in mind and referring back.

PR #395, opened 2026-03-01
fix: improve image thumbnail size to make use of available space

### LINKED ISSUE #119: Image size does not make use of available space
When the images are viewed as a grid of thumbnails, they are always shown at very small size given the space that is available. Regardless of the thumbnail size, there is always an extremely wide grey border which represents wasted space. Similarly, the information screen has a huge black border. In both cases it would be better to use the wasted space in order to show larger images.

(previously reported by email on 23/3/17)

Oyelakin-Mercy:
Fixes #119 

When viewing images as a grid of thumbnails in "edit previous votes," images were always shown at a very small size relative to the available space. Each thumbnail had an extremely wide grey border as wasted space.

**Changes:**
- Replaced fixed `height` and `aspect-ratio` on the card element with flex column, so the card grows naturally with its content
- Moved `aspect-ratio: 4/3` to the image container only, so images fill up their boxes.
- Set `object-fit: cover` and `height: 100%` on images so they fill the container without any distortion
- Replaced the excessive `100px` bottom margin with  `10px` that was creating a large grey gap below images
- Added `position: relative` to image containers so overlaid elements (footer, expand icon) anchor correctly
- For the ranking mode, moved the place label (1st, 2nd, 3rd, etc) inside the image container so it overlays the image as intended
- Updated grid size variants (`size-2`, `size-3`) to only control width to inherit the corrected layout from the base class.

Screenshot 2026-03-01 022651

Screenshot 2026-03-01 022527


ItsAbhinavM (changes requested):
@Oyelakin-Mercy thank you for your contribution. The changes made looks good, however I see that the pipeline failed. We can move forward once you fix it.

Issue #72, opened 2017-10-04
Flagging images while voting


MichaelNMaggs:
I'd like to be able to flag images when voting, so that I can come pull out or look at a selection later.  The option of several having different flags (eg of different colours) would be good.

One example would be to note copyright infringements, or images otherwise outside Commons' scope, as I go along.  I'd then be able to use the flagged list to go to Commons later and nominate those images for deletion.

Or, flags could be used for a variety of special needs within the contest, for example noting possible candidate images that might be eligible for some special prize.

leilaz:
One option would be to offer tag options to coordinators, and even jurors. I will prioritize this as normal given the other higher priority tasks.

slaporte:
We have two related features ready to be implemented in the client:

 - **Faves**: jurors can mark an entry as a fave ([/v1/juror/round/<round_id:int>/<entry_id:int>/fave](https://github.com/hatnote/montage/blob/master/montage/docs/api.md#v1jurorroundround_idintentry_idintfave)), and then get a list of their faves ([/v1/juror/faves](https://github.com/hatnote/montage/blob/master/montage/docs/api.md#v1jurorfaves)).
- **Flags**: jurors can mark an entry with a flag and submit a reason ([/v1/juror/round/<round_id:int>/<entry_id:int>/flag](https://github.com/hatnote/montage/blob/master/montage/docs/api.md#v1jurorroundround_idintentry_idintflag)), and then coordinators can view flagged entries ([/v1/admin/round/<round_id:int>/flags](https://github.com/hatnote/montage/blob/master/montage/docs/api.md#v1adminroundround_idintflags)) to determine if they should be manually disqualified ([/v1/admin/round/<round_id:int>/<entry_id:int>/disqualify](https://github.com/hatnote/montage/blob/master/montage/docs/api.md#v1adminroundround_idintentry_idintdisqualify)). The idea is to catch images that are obviously not eligible for WLM, copyright violations, etc. (I keep thinking of those weird photos of trampoline people that were submitted to Wiki Loves Monuments Iran!)

MichaelNMaggs:
Thanks: that sounds perfect

MichaelNMaggs:
Any chance of this being made available by early next week? We'd like if possible to use it in our round 2

leilaz:
@yarl do you have all the specifications you need for this feature? Let us know if we should discuss it more. Otherwise, I agree that it would be great to get to it soon. It's a very useful one for jurors and coordinators.

[...]

PR #396, opened 2026-03-01
feat: add manual disqualification for entries in rounds

### LINKED ISSUE #24: Disqualification
Coordinators should be able to disqualify photos at the start of the round. They should be able to select the reasons for disqualification (image size, uploaded by juror, or uploaded outside of the competition’s dates), and then confirm the number of affected entries before hitting submit.

### LINKED ISSUE #59: Need way to show disqualified images
There should be a way to explore which images are going to be disqualified, and why (before the disqualifications are set in stone). I have 215 disqualified images in the UK WLM17 campaign and I have no idea for what reason.

It would be useful to display the numbers that will be affected when each of the disqualification tick boxes are selected. And there should be a way to click through to see the full disqualified list.

### LINKED ISSUE #100: Manual disqualification of images
Need a way for the coordinator to review the images during later rounds and manually disqualify images that don't comply with the rules. Examples are: photo not in the correct country, photo does not indicate building number, copyright infringement, photo taken by employee of local Chapter or other sponsoring organisation etc.  These are factors that require detailed image-by-image review on Commons, making it impractical to do this for the perhaps 10,000 or more images entering the contest.  (In the UK, we do this manual step once we've whittled the number down to <200 or so).

Suggest a grid of thumbnails, an easy way of getting to Commons for each, and a simple tick box to disqualify on an individual basis - perhaps with an optional field for free text to explain the reason.

ItsAbhinavM:
# Description
Fixes #100 , Fixes #59 and Fixes #24
This PR adds a manual disqualification feature for disqualifying entries in a round during later rounds in a campaign. This is particularly useful when entries need to be reviewed for complaince issues that requires detailed inspection on Commons, such as
- Photo not in correct country
- Photo does not indicate building number
- Copyright infringement
- Photo take by employee of a local chapter or sponsoring organization ,etc.

### Features Added
- Disqualification Management Page: New page accessible via `/round/:roundId/disqualify` with a grid of image thumbnails for easy visual review
- Filter Tabs: Filter entries by All, Qualified, or Disqualified status
- Disqualify Action: Coordinators can disqualify entries with a required reason field
- Requalify Action: Ability to undo disqualification if done in error
- Status Awareness: Warning banner displayed when round is not paused, with action buttons hidden (backend requires round to be paused for modifications)

Images :
image
image
image


lgelauff (commented):
This is a preliminary set of feedback — not a full review. I did some testing and clicking around locally, but was unable to complete the full disqualify/requalify flow as the dev database has no images in the test round. Hopefully still helpful.

The core structure looks solid — the dqMap merge pattern is clean, the filter tabs and pagination are wired up correctly, and the round-paused guard is in place.

A few things to address before merging. 

** Typos in `DisqualifyPanel.vue`**

- `weight="quite"` on the grid-size buttons — should be `weight="quiet"`. The active grid size visual indicator will not work correctly.
- CSS class `diqualify-panel-controls` (missing 's') — the style rule targets `.disqualify-panel-controls`, so the controls div will be unstyled.

**Round status gate**

Can you confirm whether the disqualify/requalify buttons are visible and functional in each of these round states: inactive (before activation), active, paused, finalized, cancelled? The most important use case is being able to disqualify before activation, since removing ineligible images at that point affects how tasks are distributed to jurors.


**Disqualify reason — consider a dropdown with an "Other" option**

The reason field is currently a free-text area. Given that the most common DQ reasons are known in advance (wrong country, copyright issue, employee of sponsoring organization, etc.), a dropdown of predefined options would make coordinators faster and keep reasons consistent across rounds. "Other" in the dropdown could reveal a free-text field for cases not covered. This could also be a follow-up if you'd prefer to keep this PR focused — what do you think?

**`dq_user_id = 1` after disqualify**

After a successful disqualify, the code sets `dq_user_id = 1` as a placeholder — this will show the wrong user to other coordinators in a multi-coordinator campaign. Could you replace this with the actual authenticated user's ID, or do a targeted re-fetch of the disqualified list after the action?

**Performance on large rounds**

Has this been tested against a round with 15,000 images? `getRoundEntries` fetches all entries at once — at that scale it will be slow and may time out. The page already has pagination, so the entries endpoint could accept `page` and `per_page` params to let the backend do the slicing. Worth flagging as a known limitation if not addressed in this PR.

**No confirmation on requalify**

Disqualify requires a dialog with a mandatory reason. Requalify fires immediately on click with no confirmation. A simple confirmation step would make the two actions consistent and prevent accidental requalification.

**Hard-coded pagination strings**

`"Previous"`, `"Next"`, and `"Page X of Y"` in `RoundDisqualifyView.vue` are not using i18n. These need keys in `en.json` to be consistent with the rest of the UI.

**`router.back()` on the back button**

Could you test this: open a private/incognito window, navigate directly to `http://localhost:5173/round/1/disqualify`, then click the back button. What happens, and is that what you'd expect? Our concern is that with no prior history in the session, `router.back()` may navigate the user out of the app entirely rather than back to the round.

lgelauff:
Can you also please add a video of how you've been using it, and maybe think about some other scenarios we should consider testing?

ItsAbhinavM:
@lgelauff Thank you for your suggestion. I have made the changes accordingly.

Regarding your other doubts, this is how the disqualify page would work.

- The disqualify/requalify page will be visible all the time, but you can only update it if the round is not activated, which means you can disqualify entries before the round starts itself.
- The typos was a mistake and I have fixed it now, thanks for highlighting that :)
- I've added the disqualify drop down option now - it also provides a text box if you click "other" as a reason for disqualification.
- I haven't tested it out on rounds with 15,000 entries but have tried couple of times on rounds which has 1000s of entries and it did not lagged there.
- I've added a confirmation box for requalification of entries now.
- I've tested it on incognito, but anyways the page will only open if you are authenticated and once you are authentication you are anyways navigating from the home to its respective rounds page and then to the disqualification page, so the navigation will not be an issue here.

ItsAbhinavM:
Here is the video demonstration  of the feature.

https://github.com/user-attachments/assets/01c64e03-2bfb-4419-9f20-e23b8ce8b8a7

lgelauff:
Thank you this is helpful. Another usability issue I am realizing now thanks to the video is that it will probably be likely that the administrator coordinator knows in advance which images to delete based on copyright reasons for example. Scrolling through the images is less likely to be a pathway that they would use, but rather a search bar might be more useful, to search either by file name or uploader name. Maybe good to ask some feedback from the users who initially submitted the request based on the video recording that you submitted.

ItsAbhinavM:
> Thank you this is helpful. Another usability issue I am realizing now thanks to the video is that it will probably be likely that the administrator coordinator knows in advance which images to delete based on copyright reasons for example. Scrolling through the images is less likely to be a pathway that they would use, but rather a search bar might be more useful, to search either by file name or uploader name. Maybe good to ask some feedback from the users who initially submitted the request based on the video recording that you submitted.

@lgelauff yes that could be a good feature to have, however I'm only a developer and don't have a complete user perspective myself. Could you suggest whether this is required now or not, if yes I will include it in this PR. Otherwise we could open a discussion page for this and after discussing it with people another small PR could be added. What do you think ?

PS: I saw that you already gave your review, but I re-requested your review by mistake, please don't mind that :)

lgelauff:
I see this is actually solving a very old issue, so those users may not be active any longer. But I think that a great developer in an open source project like this would also have at least some engagement with the reporters - especially on whether it actually meets their needs. 
The challenge is that many of our jurors, coordinators are not professional issue writers, as you have noticed. So checking back in with them, especially for a more involved feature like this, will be valuable! In this case, let me see if I could get some people to give feedback instead.

Ysogo:
> > Thank you this is helpful. Another usability issue I am realizing now thanks to the video is that it will probably be likely that the administrator coordinator knows in advance which images to delete based on copyright reasons for example. Scrolling through the images is less likely to be a pathway that they would use, but rather a search bar might be more useful, to search either by file name or uploader name. Maybe good to ask some feedback from the users who initially submitted the request based on the video recording that you submitted.
> 
> @lgelauff yes that could be a good feature to have, however I'm only a developer and don't have a complete user perspective myself. Could you suggest whether this is required now or not, if yes I will include it in this PR. Otherwise we could open a discussion page for this and after discussing it with people another small PR could be added. What do you think ?
> 
> PS: I saw that you already gave your review, but I re-requested your review by mistake, please don't mind that :)

I'am an active administrator of photo contest and I fully endorse the request to have a search (or filter) functionality.

Ysogo:
It is correct that disqualification can only be performed by the contest administrator but disqualification proposal should be possible by any jury member. Actually jury members are the ones who scroll the set of photos and have a higher chance to detect that a picture is not appropriate for the contest theme. Ideally a button should be available during voting session to candidate a picture for disqualifaction. Actual disqualification should still be in the hands of the administaror when the round is closed; administrator should have a filtered view to see only "pictures proposed for disqualification" and should have two choices (a) confirm disqualification  (b) cancel disqualification proposal.

lgelauff:
@Ysogo I would propose a different workflow, but i agree with the idea behind it. In #72 there's an issue to 'flag' images for specific reasons. This could be interpreted as disqualification proposal, but depending on the category, even request for deletion may be appropriate. 
I think it comes down to the same thing - but it is less focused on the result and more on the concern. Any disqualification feature would then have to be accessible from some page where the administrator can filter for flagged images. 

Would that make sense?

Ysogo:
@lgelauff Proposal to link with #72 makes sense. Actually disqualification and deletion are slightly different concepts. A picture may not meet the criteria to be part of the contest but it can still be good to stay in Commons. The best flow, in my opinion, is that juror candidates for disqualification and later the contest-admin when confirming the disqualification should have the opportunity to insert the request for deletion in Commons (that will be executed by Commnons administrator according to community policy)

lgelauff:
> @lgelauff Proposal to link with #72 makes sense. Actually disqualification and deletion are slightly different concepts. A picture may not meet the criteria to be part of the contest but it can still be good to stay in Commons. The best flow, in my opinion, is that juror candidates for disqualification and later the contest-admin when confirming the disqualification should have the opportunity to insert the request for deletion in Commons (that will be executed by Commnons administrator according to community policy)

I think we agree!

frontend/src/components/Round/RoundDisqualifyView.vue
lgelauff, line 1: Is this file active?

lgelauff (changes requested):
Please don't forget to complete the internationalization of the new strings. I believe you need to edit qqq.json .

frontend/src/components/Round/RoundDisqualifyView.vue
ItsAbhinavM: Yes, this file was added by mistake as it was earlier present while debugging. Thank you for pointing it out. Will remove this soon.

frontend/src/views/RoundDisqualifyView.vue
```diff
@@ -0,0 +1,348 @@
...
+  }
+}
+const fetchData = async () => {
+  const roundId = route.params.roundId
+  loading.value = true
+  try {
+    const roundResponse = await adminService.getRound(roundId)
+    round.value = roundResponse.data
+
+    const entriesResponse = await adminService.getRoundEntries(roundId)
```
lgelauff, line 183: `GET /admin/round/<id>/entries` only returns entries that are **not** disqualified in this round (`CoordinatorDAO.get_round_entries` filters on `dq_reason=None`, `rdb.py:1158`). `dqMap` (line 191) is only used to label entries from that list, so a disqualified entry never gets back onto the page. This happens on every page load, in any round, not only the first time.

To spell out how disqualification works across rounds:
- It's stored **per round**, on the round entry.
- Entries disqualified in a round don't advance to the next round (`get_rating_advancing_group` skips them), and the automatic rules only run when files are imported, not when a round advances.
- So on round 2's page, entries disqualified in round 1 won't appear at all, even after this fix, because they were never part of round 2. Round 2's "Disqualified" tab should only show what was disqualified in round 2. With the current code, even those disappear on reload.

Suggested fix: build the list from both responses. `/disqualified` already includes the full entry (`dq.entry`), so those can go into `allEntries` next to the qualified ones. Its field names (`name`, `upload_user_text`, …) differ from the `img_*` ones from `/entries`.

A question for both of us: in round 2 or 3, would a coordinator expect to also see what was disqualified in earlier rounds (read-only), or is per-round what we want?

frontend/src/views/RoundDisqualifyView.vue
```diff
@@ -0,0 +1,348 @@
...
+
+const handleDisqualify = async (entry, reason) => {
+  const roundId = route.params.roundId
+  try {
+    await adminService.disqualifyEntry(roundId, entry.id, reason)
+    alertService.success($t('montage-disqualify-success'))
+
+    const index = allEntries.value.findIndex((e) => e.id === entry.id)
+    if (index !== -1) {
+      allEntries.value[index].dq_user_id = 1
```
lgelauff, line 248: This placeholder is still here from the April review. `dq_user_id` is a user id, and `1` is a real (different) user. Once comment 1 is fixed, you could reload `/disqualified` after a disqualify or requalify, so the page shows what the server actually stored. That also gets you the real reason text (the backend stores `specifically disqualified by <user> (<reason>)`).

frontend/src/components/Round/DisqualifyDialog.vue
```diff
@@ -0,0 +1,119 @@
...
+const reasonError = ref(false)
+
+const reasonOptions = [
+  { label: 'Wrong country', value: 'wrong_country' },
+  { label: 'Photo not in the correct country', value: 'incorrect_country' },
+  { label: 'Photo does not indicate building number', value: 'no_building_number' },
+  { label: 'Copyright infringement', value: 'copyright' },
+  { label: 'Photo taken by employee of local chapter or sponsor', value: 'employee_photo' },
+  { label: 'Other', value: 'other' }
+]
```
lgelauff, line 58: These reasons come straight from the examples in #100, which was written for Wiki Loves Monuments UK. In 2017 that contest had four national sub-contests (England, Scotland, Wales, Northern Ireland), each with its own upload campaign and category, and entries are listed buildings with a register number. So "not in the correct country" there most likely means *entered in the wrong national category*, and "does not indicate building number" means *the monument/heritage ID is missing*. Other contests using Montage (other WLM countries, Wiki Loves Earth, …) have the same kinds of problems under different names.

Suggestions:
- "Wrong country" and "Photo not in the correct country" are the same reason. Please keep one.
- Use more general wording, for example "Not eligible for this contest (outside the contest area / wrong category)", "Monument ID missing or incorrect", "Copyright problem", "Photographer not eligible (e.g. organizer or sponsor staff)", "Other". Store the reason in a stable form (the English label or the option `value`) rather than the translated text, so the audit log reads the same whatever language the coordinator uses.

frontend/src/components/Round/DisqualifyDialog.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+  <div class="disqualify-dialog">
+    <div class="disqualify-dialog-image">
+      <CommonsImage :image="entry" :width="400" image-class="disqualify-dialog-thumbnail" />
+    </div>
+    <div class="disqualify-dialog-content">
+      <h4>{{ entry.name }}</h4>
+      <div class="disqualify-dialog-form">
+        <label for="dq-reason">Reason for disqualification *</label>
```
lgelauff, line 9: Still hard-coded in English in this file:
- line 9: `Reason for disqualification *`
- line 13: `Select a reason`
- line 21: `Enter custom reason`
- line 25: `Please provide a reason for disqualification`
- lines 52–57: the six option labels (`Wrong country`, `Photo not in the correct country`, `Photo does not indicate building number`, `Copyright infringement`, `Photo taken by employee of local chapter or sponsor`, `Other`)

frontend/src/components/Round/DisqualifyPanel.vue
```diff
@@ -0,0 +1,255 @@
...
+              <undo-icon class="icon-small" />
+              {{ $t('montage-requalify-action') }}
+            </cdx-button>
+          </div>
+        </div>
+      </div>
+    </div>
+    <cdx-dialog
+      v-model:open="requalifyDialogOpen"
+      title="Requalify image?"
```
lgelauff, line 88: Still hard-coded in English in this file:
- line 42: `DQ`
- line 88: `Requalify image?`
- line 89: `Requalify`
- line 90: `Cancel`
- line 95: `Are you sure you want to requalify "…"?`

frontend/src/views/RoundDisqualifyView.vue
```diff
@@ -0,0 +1,348 @@
...
+        @requalify="handleRequalify"
+      />
+      <div v-if="totalPages > 1" class="disqualify-pagination">
+        <cdx-button :disabled="currentPage === 1" @click="prevPage">
+          <chevron-left class="icon-small" />
+          Previous
+        </cdx-button>
+        <span class="disqualify-page-info"> Page {{ currentPage }} of {{ totalPages }} </span>
+        <cdx-button :disabled="currentPage === totalPages" @click="nextPage">
+          Next
```
lgelauff, line 82: Still hard-coded in English:
- line 78: `Previous`
- line 80: `Page {{ currentPage }} of {{ totalPages }}`
- line 82: `Next`

frontend/src/views/RoundDisqualifyView.vue
```diff
@@ -0,0 +1,348 @@
...
+    <div v-if="loading" class="disqualify-loading">
+      <p>{{ $t('montage-disqualify-loading') }}</p>
+    </div>
+
+    <div v-else-if="round">
+      <div v-if="round.status !== 'paused'" class="disqualify-warning">
+        <alert-icon class="disqualify-warning-icon" />
+        <p>
+          {{ $t('montage-disqualify-readonly-warning') }}
+          <strong>{{ round.status }}</strong>
```
lgelauff, line 21: Minor: this shows the raw status value (`active`, `finalized`, …) inside a translated sentence. Fine to leave for a follow-up.

lgelauff (changes requested):
Thanks for sticking with this one. The qqq.json strings are in, and the search by file name and uploader is really useful. Coordinators often already know which files they want to remove.

**Disqualified entries seem to disappear after a reload**

I asked Claude to test this against the database in my local setup, and the result suggests that disqualified entries disappear from the page after a reload. This is what it did, step by step:

1. Start from a round with 20 entries and no disqualifications.
2. Pause the round.
3. Disqualify one entry (the same `POST /admin/round/<id>/<entry_id>/disqualify` call the page makes).
4. Load the page data again the way the page does on load: `GET /admin/round/<id>/entries` and `GET /admin/round/<id>/disqualified`.
5. Result: `/disqualified` returns the entry, but `/entries` now returns 19 entries and doesn't include it. The page shows 0 disqualified entries and 19 in "All".

In the UI this would mean: disqualify an image, reload the page, and the image is gone. It isn't under "Disqualified", so it can't be requalified from here either. I've left more detail inline.

Have you tested this locally? If you follow the same steps, do you see the same thing?

**Follow-up issue**

Once this is merged, could you open a new issue for the parts of #100 and #24 that this doesn't cover yet? This is what I think is still open:

- From #100: disqualifying several images in one go. #100 asks for a tick box per image, while this PR asks for a reason one image at a time. For the "couple of hundred images" case in #100, that means a lot of dialogs.
- From #24: all of it. When starting a round, choose which automatic disqualification rules apply (image size, uploaded by a juror, uploaded outside the contest dates), see how many entries each rule would remove, and confirm before the round starts. (The preview of counts is also what #59 asks for.)

**Performance on large rounds**

I'm not sure how this holds up on a big round. The page loads every entry at once. Could you try it on a category with 10,000 or 30,000 images and let us know what opening the page looks like?

If it turns out to be too slow, do you have a suggestion for getting around it? One option I could see: don't build the list on the spot. Tell the coordinator it will take a while and to come back in an hour or so, store the result, and show it on the page once it's ready, rather than as a pop-up.

PR #397, opened 2026-03-02
feat: add share vote with consent on a campaign

### LINKED ISSUE #150: Option to share voting dataset with consent
It may be useful to give researchers open access to anonymized sets of votes. We can add an option for jurors (and coordinators) to opt-in to making their voting information public.

Features:
 - [ ] Opt-in flag for round jurors
 - [ ] Opt-in flag for campaigns set by coordinators
 - [ ] Page describing datasets

ItsAbhinavM:
# Description
Fixes #150 
This PR implements a feature that allows campaign coordinators to opt-in to sharing anonymized voting data for research purposes.
Changes made:
- Added coordinator opt-in toggle for research data sharing when creating or editing campaigns
- Created a public datasets page where anyone can view and download anonymized voting data from opted-in campaigns
- Export data is minimal and privacy-focused: only filename and vote value (no juror information)


lgelauff:
I would have expected a migration file and a test for the new endpoint. Is there a reason why this should not be added?

PR #410, opened 2026-03-15
Fix/preemptive checks round creation#103

Sanika123-art:
### Summary

This pull request adds pre-emptive validation checks to prevent disallowed user actions when creating a round.

Previously, certain invalid inputs (such as missing jurors, invalid quorum values, or invalid Wikimedia Commons categories) could cause the backend to return a 500 Internal Server Error. This PR adds validation logic so that these cases are handled gracefully and return appropriate error messages instead of crashing the server.

### Changes Made

* Added validation for missing jurors before creating a round.
* Added validation for invalid quorum values (quorum must be greater than 0).
* Added validation for invalid or empty Wikimedia Commons categories during image import.
* Added proper error handling to return HTTP 400 responses instead of causing 500 server errors.

### Expected Behavior

* If a user attempts to create a round without jurors, the system returns a validation error.
* If quorum is set to an invalid value, the request is rejected.
* If an invalid category is provided for image import, the backend returns a clear error message instead of crashing.

### Result

These checks prevent backend crashes and improve system stability by ensuring invalid user actions are caught before processing.


Sanika123-art:
Hi! I have resolved the merge conflicts and updated the branch.
Could you please review the PR when you have time? Thank you!

PR #412, opened 2026-03-17
Fix: Improve format_date function to handle None and invalid datesImprove error handling in format_date function

Harshita-kothari:
### Summary
This PR improves the `format_date` function in utils.py to safely handle cases where the date value is None or invalid.

### Changes
- Added a check for None before formatting the date
- Wrapped date formatting in a try/except block to prevent crashes

### Why this change
Previously, the function could raise an exception if the date was None or invalid. This fix ensures the function fails gracefully by returning an empty string.

### Testing
- Tested with None values
- Tested with valid datetime values

### DIFF
--- a/montage/utils.py
+++ b/montage/utils.py
@@ -198,9 +198,13 @@ def check_schema(db_url, base_type, echo=False, autoexit=False):
 
 
 def format_date(date):
-    if isinstance(date, datetime.datetime):
-        date = date.isoformat()
-    return date
+    if date is None:
+        return ""
+
+    try:
+        return date.strftime("%Y-%m-%d")
+    except Exception:
+        return ""
 
 
 def js_isoparse(date_str):


lgelauff (commented):
The PR introduces a **silent breaking change**: `format_date` previously returned`.isoformat()` (e.g. `"2023-06-15T14:30:00"`), but now returns`strftime("%Y-%m-%d")` (e.g. `"2023-06-15"`). This changes the API response format for every date field across all models

**Critical bug — campaign edit form crashes:**
`ViewCampaign.vue:243,245` does `open_date?.split('T')[1].substring(0, 5)` to populate the time fields when opening the edit modal. With the new date-only format, `split('T')[1]` is `undefined` and `.substring()` throws a `TypeError`. Every campaign edit attempt will crash.

**Secondary — 1-day-off in negative-UTC timezones:** `RoundInfo.vue:160` passes `deadline_date` directly to `new Date()`. A date-only string (no time/zone) is parsed as UTC midnight, which renders the previous calendar day for users in UTC− timezones (e.g. US).

**Safe fix:** keep `.isoformat()`, just add the `None` guard:
```python
def format_date(date):
    if date is None:
        return None
    if isinstance(date, datetime.datetime):
        return date.isoformat()
    return date ```

but this would need more testing. Converting to draft

lgelauff:
For now, I'll close this. Feel free to open a new PR referencing this later.

PR #418, opened 2026-03-21
Fixes #322 Edit votes loads only 15 images (scroll listener race condition)

### LINKED ISSUE #322: 'Edit votes' loads very slowly or does not display images at all
'Edit previous photos' window loads very slowly, sometimes it loads the last 15 photos only and simply fails to load any of the rest. That was not the case in the old frontend.

ayushshukla1807:
Fixes #322.

Edit Votes was dropping the queue after 15 images because the intersection observer was firing out of order. Adjusted the listener hook.


ayushshukla1807:
I am closing this PR to reduce repository noise. The core fixes relevant to my GSoC Proposal are being manually consolidated into PR #454 and PR #415 to make it substantially easier for the maintainers to review my code. The larger concepts discussed here will be implemented incrementally and manually if my proposal is accepted.

ayushshukla1807:
### 🧪 Local Testing Verification

Verified locally. The scroll-listener race condition has been patched. The Vue component is correctly detecting the scroll depth and firing the subsequent pagination network requests.

As shown below, scrolling down the gallery correctly loads the 16th+ images without hanging.

PR #421, opened 2026-03-22
Automate setuptools installation

harshithaMMPS333:
This pull request  automates the **setuptools**(third party library) installation.

Instead of installing it manually now tox automatically does it,

Added **setuptools** in deps- tox.ini

Before:- 
Screenshot 2026-03-22 at 5 15 07 PM
**without manually installing setuptools** 

 After:-
Screenshot 2026-03-22 at 5 17 19 PM

as of now there is no requirement to install it manually.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,11 @@
+#To install backend packages, run this command in CLI
+# > python3 -m venv venv
+# > source venv/bin/activate
+# > pip install -r requirements-dev.txt 
 -r requirements.txt
 
-tox<3.15.0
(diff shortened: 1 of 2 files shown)


lgelauff:
FYI: we've added a `setuptools==81.0.0` pin to `requirements.txt` as a workaround for setuptools 82 removing `pkg_resources` — which `python-graph-core` depends on via namespace package declarations. When this PR (or a follow-up) removes the `python-graph-core` dependency or migrates away from `pkg_resources`, please also remove the setuptools pin from `requirements.txt`.

PR #432, opened 2026-03-27
fix: add missing disabled binding to Close Campaign button

MuraariK13052008:
## What this fixes
The "Close Campaign" button was always clickable regardless of whether the campaign was actually active.

## Root cause
`canCloseCampaign` is a reactive ref that is correctly set to `true` only when `campaign.status === 'active'` inside `reloadState()`. However, the Close Campaign button in the template had no `:disabled` binding at all — the ref was declared and populated but never consumed in the template, making the button permanently enabled.

## Fix
Added `:disabled="!canCloseCampaign"` to the Close Campaign button on line 7 of ViewCampaign.vue.

## Files changed
- `frontend/src/components/Campaign/ViewCampaign.vue`

## How to verify
1. Open a campaign that is not active (e.g. finalized or paused)
2. Confirm the "Close Campaign" button is now disabled
3. Open an active campaign — confirm the button is enabled

Relates to: T415578

### DIFF
--- a/frontend/src/components/Campaign/ViewCampaign.vue
+++ b/frontend/src/components/Campaign/ViewCampaign.vue
@@ -4,7 +4,7 @@
       <div class="campaign-header">
         <p class="campaign-title">{{ campaign.name }}</p>
         <div class="campaign-button-group">
-          <cdx-button action="destructive" weight="primary" @click="closeCampaign">
+          <cdx-button action="destructive" weight="primary" @click="closeCampaign" :disabled="!canCloseCampaign">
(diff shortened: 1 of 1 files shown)


ItsAbhinavM:
@MuraariK13052008 thank you for your contribution. I've tested your PR and it looks good, I also noticed that you fixed the typo of `isArchive` and `data-testId` which is great, since it is a small change it is fine to keep it here but could have briefed in the description.

@lgelauff I think we are good to go with this PR, could you please do a quick check before we finalize it ?

ItsAbhinavM (approved):

PR #435, opened 2026-03-27
fix: replace router.reload with location.reload

MuraariK13052008:
## What this fixes
After deleting a round, the page never refreshed — the deleted round remained visible in the UI even though it had been removed from the database.

## Root cause
Line 194 of RoundEdit.vue calls `router.reload()` after a successful round deletion. Vue Router does not have a `reload()` method — this throws `TypeError: router.reload is not a function` at runtime. Because the error was thrown inside a `.then()` callback, it was silently swallowed and the page never refreshed.

## Fix
Changed `router.reload()` → `location.reload()`, which is the correct browser API for a full page reload. This is consistent with the exact same pattern already used in RoundInfo.vue for `activateRound()` and `pauseRound()`.

## Files changed
- `frontend/src/components/Round/RoundEdit.vue`

## How to verify
1. Open any campaign with at least one round
2. Delete a round using the delete action
3. Confirm the page now refreshes automatically and the deleted round no longer appears

Relates to: T415578

### DIFF
--- a/frontend/src/components/Round/RoundEdit.vue
+++ b/frontend/src/components/Round/RoundEdit.vue
@@ -191,7 +191,7 @@ const deleteRound = () => {
         .cancelRound(props.round.id)
         .then(() => {
           emit('update:isRoundEditing', false)
-          router.reload()
+          location.reload()
         })
         .catch(alertService.error)
     }


lgelauff:
Thanks for submitting. It looks like https://phabricator.wikimedia.org/T415578 if your own ubmrella task. Have you looked for issues on Github that this would be relevant to?

lgelauff (commented):
Correct fix, verified locally — deleting a round now refreshes the page as expected.

One note: the PR references T415578 (the GSoC umbrella task) but not GitHub issue #379, which directly describes this bug. Please add "Fixes #379" to the PR description.

lgelauff:
I'll leave to @mahmoud / @Jayprakash-SE if the reload is too blunt an instrument here, or that a solution closer to `reloadCampaignState` is better

ItsAbhinavM:
I believe this issue has already been solved in #380 which has been merged.

lgelauff:
indeed. thanks. closing as solved.

Issue #370, opened 2026-02-03
Permission problem for Jurors


zacharyb02:
Hello,
For the Wiki Loves Clermont-Ferrand campaign, I created a Yes/No round. However, 3 out of 4 jurors encountered a "Permission denied" message, even though their usernames are in the jurors list and the round is active.

<img width="226" height="84" alt="Image" src="https://github.com/user-attachments/assets/94e8e11b-beba-4989-9dca-9ca6daaa1aea" />

Santhini16:
Hello @zacharyb02 , I’m a Santhini , A contributor interested in working on this issue.

I’d like to investigate the permission problem for jurors. Could you please guide me on where the permission logic is handled in the codebase?

Thank you!

Santhini16:
Hi, I have submitted a pull request to fix this issue. Please review it and let me know if any changes are required. Thank you!

lgelauff:
> Hello, For the Wiki Loves Clermont-Ferrand campaign, I created a Yes/No round. However, 3 out of 4 jurors encountered a "Permission denied" message, even though their usernames are in the jurors list and the round is active.
[...]

PR #441, opened 2026-03-29
Fix juror permission issue (#370)

Santhini16:
#### Problem

Jurors were encountering a "Permission denied" error when attempting to access rounds or campaigns, even though:

* Their usernames were correctly listed as jurors
* The round was active and properly configured

The issue was caused by the permission check relying on username-based matching:

```python
Round.jurors.any(username=self.user.username)
```

Using `username` for permission checks can become problematic if a user's username changes, while `user.id` remains stable. Using the primary key (`id`) is therefore a more reliable approach for permission validation.

#### Solution

Updated the permission validation logic to use the user's unique identifier (`id`) instead of username for comparison:

```python
Round.jurors.any(id=self.user.id)
```

Additionally, `_get_round_juror` was updated to consistently use `user_id`:

```python
.filter_by(user_id=self.user.id, round_id=round_id)
```

**Impact**

* Ensures accurate and reliable permission checks for jurors
* Eliminates false "Permission denied" errors
* Aligns permission logic with best practices (using primary keys instead of string fields)

**Scope of Changes**

* Modified `get_campaign` and `get_round` methods in `JurorDAO`
* Updated `_get_round_juror` for consistency

 **Notes**

This change is minimal and does not affect other query logic or relationships. Existing functionality remains intact while improving correctness of access control.

Please let me know if any further adjustments or tests are required.

### DIFF
--- a/montage/rdb.py
+++ b/montage/rdb.py
@@ -2473,7 +2473,7 @@ class JurorDAO(object):
             return self.user_dao._get_any_campaign(campaign_id)
         campaign = self.query(Campaign)\
                        .filter(Campaign.rounds.any(
-                           Round.jurors.any(username=self.user.username)))\
+                           Round.jurors.any(id=self.user.id)))\
                        .filter_by(id=campaign_id)\
                        .one_or_none()
         if not campaign:
@@ -2485,7 +2485,7 @@ class JurorDAO(object):
             return self.user_dao._get_any_round(round_id)
         rnd = self.query(Round)\
                   .filter(
-                      Round.jurors.any(username=self.user.username),
+                      Round.jurors.any(id=self.user.id),
                       Round.id == round_id)\
                   .one_or_none()
         if not rnd:


Santhini16:
Hi, I have submitted a pull request to fix this issue. Please review it and let me know if any changes are required. Thank you!

ItsAbhinavM:
@Santhini16  thank you for your contribution, I do believe that the main reason of the bug is because of the comparison with username than ID. If the username is changed mid campaign then this will raise an error and an ID would fix that issue.
@lgelauff Could you please do a quick review before merging ? I've reviewed, tested locally and its looks good :)

Santhini16:
@ItsAbhinavM Thank you for the review and for testing the fix locally! I appreciate it really

lgelauff:
> @Santhini16 thank you for your contribution, I do believe that the main reason of the bug is because of the comparison with username than ID. If the username is changed mid campaign then this will raise an error and an ID would fix that issue. @lgelauff Could you please do a quick review before merging ? I've reviewed, tested locally and its looks good :)

@Santhini16 you indicate that it's unreliable. Were you able to reproduce it at least for a set of cases? If you like, I can give you access to montage-beta for trying to reproduce with a live user db with mediawiki accounts. 

I suspect that local testing is unreliable here because of the authentication integration.

Santhini16:
@lgelauff Thank you for the feedback. You're right ,I wasn't able to reliably reproduce the issue locally, especially because the authentication integration is difficult to simulate in my development environment. My change was based on the code path and the observation that using user.id is a more stable identifier than username for permission checks. I'd be happy to test this further on montage-beta if you can provide access.

ItsAbhinavM:
I think if you want to test the username credibility, try changing it on the local `tmp_montage.db` and see whether you could reproduce the bug.

PR #454, opened 2026-03-30
Fix: Prevent 500 POST errors via strict frontend schema validation during Round Creation

### LINKED ISSUE #447: First round creation without images succeeds with zero entries, blocking subsequent round creation

ayushshukla1807:
Fixes #447.

Tightened the frontend constraints on `RoundNew.vue` so people can't submit broken or empty round configurations and trigger unhandled 500 stack traces on the backend.

### DIFF
--- a/frontend/src/components/Round/RoundNew.vue
+++ b/frontend/src/components/Round/RoundNew.vue
@@ -312,17 +312,40 @@ const cancelRound = () => {
 
 const submitRound = () => {
   if (!formData.value.deadline_date) {
-    alertService.error({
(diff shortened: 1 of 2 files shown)


ayushshukla1807:
### 🧪 Local Testing Verification

I've successfully verified this fix locally. As shown in the screenshot below, when an organizer attempts to submit an empty form, the synchronous frontend validation intercepts the payload **before** the network request is fired.

This immediately resolves the `KeyError`/`ValueError` 500 internal server crash (Issue #447/#451) and provides instant UI feedback to the organizer without wasting backend resources.

Issue #462, opened 2026-04-01
[Bug]: `finalize_round` endpoint bypasses DAO finalization logic (missing close_date, audit log)


userAdityaa:
## Description:
The finalize_round endpoint at `admin_endpoints.py` sets `rnd.status = FINALIZED_STATUS` directly on the ORM object instead of calling the proper DAO methods (**finalize_rating_round** or **finalize_ranking_round**). This means:

* `close_date` is never set on the round
* Update status to finalized
* No audit log entry is created

## Steps to Reproduce:

1. Create a campaign, add a round, activate it, and complete all voting
2. Finalize the round via the endpoint
3. Check the round's `close_date` in the database, it is NULL
4. Check the audit log, no `finalize_round` entry exists

## Expected Behavior

Finalizing a round should perform all required finalization actions, including:

* Set `close_date` to the current timestamp
* Create an audit log entry for the finalization action
* For **ranking rounds**: create a `RoundResultsSummary`
* For **rating rounds**: store the `final_threshold` in the configuration

## Actual Behavior

Only the status field is changed. All other finalization side-effects are skipped.

lgelauff:
The bug is real, but the scope is narrower than described.

`finalize_round` is a standalone "close this round" action — it is not responsible for computing the advancing group or storing the threshold. Those are determined at the point of creating the *next* round via `advance_round`, where the coordinator selects a threshold and the advancing entries are calculated. The `final_threshold` stored in the closing round's config is a side effect of that decision, not something `finalize_round` should set. In fact, the threshold is more associated with starting the next round, than closing the current: if we just close, no threshold is needed. 

The actual fix needed is:
- Set `close_date`
- Update status to finalized
- Write an audit log entry (to be verified)

No threshold logic, no result summaries. Those belong in `advance_round` and `finalize_campaign` respectively, where they already live.

ayushshukla1807:
Got it, that scope makes sense for result summaries and thresholds handling. My PR #466 routes it directly to coord_dao.finalize_round to catch the close_date properly. Appreciate the heads up!

PR #463, opened 2026-04-01
fix: finalize_round: delegate to proper DAO methods instead of bare s…

userAdityaa:
…tatus update

## Summary

Fix the `finalize_round` endpoint in `admin_endpoints.py` to ensure all required finalization actions are executed by delegating to the appropriate DAO methods instead of only updating the round status.

## Solution

The endpoint now properly delegates round finalization logic to the DAO layer:

* Calls `finalize_rating_round` for rating rounds
* Calls `finalize_ranking_round` for ranking rounds
* Ensures all expected finalization side-effects are executed

**Fixes** #462 

## Changes

* Call DAO finalization methods instead of directly updating status
* Ensure `close_date` is set during finalization
* Ensure audit log entries are created
* Ensure ranking rounds create `RoundResultsSummary`
* Ensure rating rounds store `final_threshold` in config
* Remove unused `FINALIZED_STATUS` import

## How to test

1. Create a campaign and add a round
2. Activate the round and complete voting
3. Finalize the round via the endpoint
4. Verify:

   * `close_date` is set in the database
   * Audit log entry exists for finalization
   * Ranking rounds create a `RoundResultsSummary`
   * Rating rounds store `final_threshold` in config

### DIFF
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -14,8 +14,7 @@ from .utils import (format_date,
                    NotImplementedResponse,
                    js_isoparse)
 
-from .rdb import (FINALIZED_STATUS,
-                 CoordinatorDAO,
+from .rdb import (CoordinatorDAO,
                  MaintainerDAO,
                  OrganizerDAO)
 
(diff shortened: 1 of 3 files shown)


lgelauff:
Minor nit: the test fixture uses `'Images_from_Wiki_Loves_Monuments_2015_in_albania'` (lowercase `a` in `albania`), but the actual Commons category is `Images_from_Wiki_Loves_Monuments_2015_in_Albania` (capital `A`). Worth correcting to match the real category name, even in tests.

userAdityaa:
> Minor nit: the test fixture uses `'Images_from_Wiki_Loves_Monuments_2015_in_albania'` (lowercase `a` in `albania`), but the actual Commons category is `Images_from_Wiki_Loves_Monuments_2015_in_Albania` (capital `A`). Worth correcting to match the real category name, even in tests.

Yes, that make sense. Thanks for pointing that out.

Pushing changes in a bit.

userAdityaa:
Resolved the merge conflicts and addressed the test refactoring based on the feedback. Thanks!

lgelauff:
Looking more carefully, I believe the scope of the issue #462 was incorrect, and this PR is based on some assumptions from that issue. 

I'm converting to draft, but actually would recommend closing this and possibly starting a new PR to keep it clean. 

PR #463 delegates to `finalize_rating_round` and `finalize_ranking_round`, which do more than `finalize_round` should — and introduces a required `threshold` parameter that breaks the existing frontend call. The fix should be simpler: a new DAO method that cleanly closes the round.

userAdityaa:
Hey @lgelauff, Thanks for the careful review. I rechecked the flow and agree that my solution expanded scope too much.

I’ll clean this up, but I can also rewrite the commit history within the same PR so the review feedback is preserved. Otherwise, yes, opening a new PR would be the cleaner option.

Issue #464, opened 2026-04-03
[Bug]: finalizeRound API path has leading slash, bypassing baseURL prefix


userAdityaa:
### Description
In `adminService.js:48`, the `finalizeRound` endpoint has a leading `/` that other endpoints don't:

All other endpoints use relative paths like:
- `admin/round/${id}/cancel`
- `admin/round/${id}/activate`

With axios, when a URL starts with `/`, it is resolved as absolute relative to the origin, ignoring the `baseURL` (which includes `/v1/`).

### Steps to Reproduce

1. As a coordinator, attempt to finalize a round.
2. Observe the network request goes to `/admin/round/{id}/finalize` instead of `/v1/admin/round/{id}/finalize`.

### Expected Behavior

Request should go to:

```js
{baseURL}/admin/round/{id}/finalize
```
(i.e., `/v1/admin/round/{id}/finalize`).

### Actual Behavior

Request goes to:

```js
/admin/round/{id}/finalize
```

(missing `/v1/` prefix), resulting in a **404 or routing error**. Round finalization fails from the frontend.

lgelauff:
This took some digging, because also my Clause got confused over this. However, it seems that the leading slash is a style inconsistency. I don't observe the behavior described in production (or locally). Axios seems to handle it fine. Please make sure to verify a bug actually exists by at least locally verifying it in person. Closing the issue.

PR #465, opened 2026-04-03
Style: Normalize backend endpoint URL slashes for Axios consistency

ayushshukla1807:
Quick style cleanup.

Forced explicit trailing slashes onto the Axios base URLs so we aren't wasting time relying on the native router doing auto-301 redirects.

### DIFF
--- a/frontend/src/services/adminService.js
+++ b/frontend/src/services/adminService.js
@@ -49,7 +49,7 @@ const adminService = {
 
   advanceRound: (id, data) => apiBackend.post(`admin/round/${id}/advance`, data),
 
-  finalizeRound: (id) => apiBackend.post(`/admin/round/${id}/finalize`),
+  finalizeRound: (id) => apiBackend.post(`admin/round/${id}/finalize`),
 
   // Direct download URLs (manual baseURL needed)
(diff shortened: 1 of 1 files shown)


userAdityaa:
Hi @ayushshukla1807, thanks for opening the PR. Could you please add a short description explaining the change and what behavior we should expect after it? This will make it much clearer for future contributors.

ayushshukla1807:
Thanks for the suggestion, @userAdityaa. To clarify context for future contributors: this patch addresses a regression where the redundant leading slash bypassed the global Axios `baseURL` interceptor (`/v1`). By removing it, the `finalizeRound` payload correctly inherits the `v1` prefix routing automatically.

userAdityaa:
Hey, I meant updating the PR description itself rather than adding the explanation only in the comments.

ayushshukla1807:
Done! The description has been updated with the technical context.

lgelauff:
Thanks for looking into this. It seems impossible to reproduce this error ( #464 ) locally. It's not a big fix in that sense, but perhaps a style cleanup. I'll leave it to others to determine whether that is desirable. Please adjust the description and title accordingly. I'll move it to draft in the mean while.

PR #467, opened 2026-04-04
fix(rounds): show and inherit DQ flags when creating new rounds

### LINKED ISSUE #347: Flag setting for disqualified file can not be set when creating Round 2
When creating round 2, the disqualification flags are not available for settings and (even worse) it is not inheriting settings from Round 1 but it decides on its own.

Image

Can you please either:
- make flags available for settings with starting values from previous Round (best solution)
- make flags available for settings without any preset                                     (good solution)
- keep no edit logic but use the same setting of Round 1                               (fair solution)

Thank you

userAdityaa:
**Description**  
This PR ensures the disqualification (DQ) flags panel is always visible when creating any round, inherits settings from the previous round for Round 2 and beyond, sends the correct config to the backend, and adds the missing `dq_by_filetype` option.

**Solution**  
- Always render DQ flags panel for all rounds  
- Inherit config from previous round for new rounds  
- Include config in `advance_round` API payload  
- Add `dq_by_filetype` to file settings

Fixes #347 

**How to Test**  
1. Create Round 1: DQ flags panel is visible and editable  
2. Create Round 2+: DQ flags panel is visible, pre-filled with previous values, and editable  
3. Confirm new round uses selected DQ settings

**Screenshots**  
|Before|After|
|----|----|
|Screenshot 2026-04-04 at 3 36 14 PM|Screenshot 2026-04-04 at 3 36 57 PM

### DIFF
--- a/frontend/src/components/Round/RoundNew.vue
+++ b/frontend/src/components/Round/RoundNew.vue
@@ -1,12 +1,27 @@
 <template>
   <div class="juror-campaign-round-card">
     <div class="round-header">
-      <thumbs-up-down v-if="formData.vote_method === 'yesno'" class="juror-campaign-round-icon" :size="36"
-        fillColor="white" style="background-color: grey" />
-      <star-outline v-if="formData.vote_method === 'rating'" class="juror-campaign-round-icon" :size="36"
-        fillColor="white" style="background-color: grey" />
-      <sort v-if="formData.vote_method === 'ranking'" class="juror-campaign-round-icon" :size="36" fillColor="white"
-        style="background-color: grey" />
+      <thumbs-up-down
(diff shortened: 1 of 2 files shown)


lgelauff:
Could you please review your PR based on my comment on #347 ? Thanks. Converting to draft in the mean while.

userAdityaa:
> Could you please review your PR based on my comment on #347 ? Thanks. Converting to draft in the mean while.

Hey @lgelauff, I’ve opened the PR for review. I’ve tested everything on my end, and the fix looks solid, ready for review and merge when you’re good with it.

the test is included in a seperate atomic commits and follows all scenario as per https://github.com/hatnote/montage/issues/347#issuecomment-4229467383

Issue #324, opened 2025-10-04
Odd behavior during file import


atsirlin:
Importing images produces an optimistic message 'Round successfully created', along with an empty round that returns an error upon pressing the 'Activate' button. Only after a delay of 2-3 min do the photos appear, but during this period it's impossible to see if they are being imported, or an error occurred during the import.

lgelauff:
I was unable to reproduce this problem myself. 

Could you please describe step by step how you imported the round, and at what time? Am I correct in understanding that eventually it worked? If the process eventually works out, that is good to know - because while it's suboptimal to get confusing error messages, it's limited to only coordinators and not jurors.

ademolaomosanya:
i was able to reproduce the bug while importing "Images from Wiki loves Monuments 2025 in Belgium" on my local setup

<img width="1440" height="659" alt="Image" src="https://github.com/user-attachments/assets/6ce0909a-abd2-45ea-85d5-6f00e9688193" />

My proposed fix will be to  handle import errors while the process is running in the background on the server before displaying the success message to the user. a completion message will only be shown after the process finishes, allowing us to return the appropriate feedback to the user and address any delays in server file import through optimization.
i'm working on this issue and will raise a pull request for it.

PR #469, opened 2026-04-05
Fix categorylinks query for MediaWiki link target normalization

### LINKED ISSUE #425: Round created, but without images
While attempting to create a new round, seems to be happening some db error that prevents images to be loaded.

Traceback:
Internal server error: <ExceptionInfo [pymysql.err.OperationalError: (1054, "Unknown column 'cl_to' in 'ON'")] (49 frames, last=Callpoint('raise_mysql_exception', 150, 'pymysql.err', '/data/project/montage/www/python/venv/lib/python3.9/site-packages/pymysql/err.py', 120, '    raise errorclass(errno, errval)'))>

lgelauff:
## Summary

- **labs.py**: replace `cl_to` with a join on the new `linktarget` table (`cl_target_id = lt_id AND lt_namespace = 14 AND lt_title = %s`), following Wikimedia's link target normalization schema change (2022-2023)

## Notes

- The `cl_to` column in `categorylinks` was replaced by `cl_target_id` (FK to `linktarget`) on Wikimedia replicas. This broke category import on Toolforge.
- The `xfail` marks originally in this PR have been dropped — master now has properly mocked tests covering the same cases.
- Verified working on montage-beta: category import completes successfully.

## Test plan

- [x] Deploy to montage-beta and verify category import works
- [x] xfail marks removed (superseded by mocked tests in master)
- [x] CI passes (6/6 tests passing)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,7 @@ cypress/screenshots/
 cypress/videos/
 
 .vscode/
+
+# Claude Code
+.claude/
+tmp/
--- a/montage/labs.py
+++ b/montage/labs.py
@@ -72,7 +72,9 @@ def get_files(category_name):
         AND page_title = img_name
         JOIN categorylinks ON cl_from = page_id
         AND cl_type = 'file'
-        AND cl_to = %s
+        JOIN linktarget ON cl_target_id = lt_id
+        AND lt_namespace = 14
+        AND lt_title = %s
         GROUP BY img_name
         ORDER BY oi_timestamp ASC;
     '''.format(cols=', '.join(IMAGE_COLS))


lgelauff:
Currently the CI fails also on Master because of this error. Had to do a workaround for passing the tests. I created #470 to change the test suite to be no longer dependent on the live API.

lgelauff:
Tested the fix from PR #469 on montage-beta. Category import now works — the round is created and files are imported correctly. There is a UX delay where no feedback is shown while the import runs in the background (tracked separately in #324), but the underlying import works.

mahmoud (approved):
LGTM, wish we had a better approach to testing, but we didn't before so no need to block now when the service is broken. Thank you!

PR #471, opened 2026-04-05
Frontend/round validation 

SparshM8:


lgelauff:
Thanks for working on this. Unfortunately there is no video of the fixes in action, and the documentation of what you have changes is a bit on the brief side. 

- We will converge on playwright as our testing approach - not your fault, but we won't be including the cypress testing for now. 
- Please follow the `i18n` framework. I improved documentation in #530 (coming up). 
- The helper functions seem useful and clean. Were you able to see them in action in a local deployment? 
- Lets extend the pattern with zod from `NewCampaign.vue` and `ViewCampaign.vue`. But given the overlap with #381 perhaps good to coordinate. 

Converting to draft for now, but definitely still useful elements, I think!

PR #476, opened 2026-04-09
Fix category import 500

### LINKED ISSUE #447: First round creation without images succeeds with zero entries, blocking subsequent round creation

Sanika123-art:
**####  ## Problem**

When creating a round using the category import method, invalid or missing category input caused unhandled exceptions in:

```
coord_dao.add_entries_from_cat()
```

This resulted in a **500 Internal Server Error**, breaking the backend flow and preventing further processing like entry handling and disqualification.

---

####  Solution

* Added proper validation for category input
* Ensured `Category:` prefix is correctly applied
* Wrapped category import logic in a `try-except` block to handle failures safely
* Added meaningful warning messages instead of allowing crashes
* Handled edge cases:

  * Missing category
  * Invalid category
  * No images returned

---

####  Key Changes

* Prevented backend crash by catching exceptions during category import
* Ensured `entries` is always defined (empty list fallback)
* Added user-friendly warnings:

  * `Category is required`
  * `No images found in this category`
  * `Invalid category or unable to fetch images`
* Allowed downstream logic (like disqualification) to execute correctly

---

####  Result

*  No more 500 Internal Server Errors
*  Backend flow continues even on invalid input
*  Warnings are shown properly to the user
*  Disqualification and further processing work correctly

---

####  Testing

* Tested with empty category → shows warning
* Tested with invalid category → no crash, warning shown
* Tested with valid category → entries imported successfully
* Verified disqualification logic still executes

---

Closes #447


Sanika123-art:
Hi! Adding a quick note for context:

This PR focuses on fixing the 500 Internal Server Error (#447) by handling invalid or missing category input at the backend level. The goal is to prevent crashes and ensure the flow continues with proper warnings.

I also noticed there is a frontend validation approach being discussed. I agree that preventing invalid requests at the UI level is valuable, and combining both frontend validation and backend safeguards would make the system more robust.

I’ve also resolved the merge conflicts and ensured the changes are cleanly integrated.

Happy to make any further improvements if needed. Would really appreciate your feedback!

@ayushshukla1807 would like your thoughts on this approach as well!

lgelauff:
Please review my comment on #447 - updates in the master have provided new insights into the required scope. Converting back to draft for now.

ItsAbhinavM (commented):
@Sanika123-art thank you for your contribution. The issue #447 mainly targets about the problem with submitting null entries to the backend which later turns into a bug leading to round creation issue. I went through the changes and felt that the changes made does not quite fulfills the intended target of the issue. 

The PR have some unrelated changes on `admin_endpoints`, `rdb` files and the changes on the frontend are validations on the parameters which doesn't cause the issue. Please feel free to raise a new PR if you think you can provide a better approach on fixing this.

ItsAbhinavM:
Closing this PR as we are not moving forward with this approach.

PR #480, opened 2026-04-10
Fix: Surface import failures properly to frontend instead of throwing CORS 500

### LINKED ISSUE #475: Images not being imported for new rounds
# Description

When you try to create new rounds on a campaign, the entries are not being imported from commons. The API which is failing is `/admin/round/<round_id:int>/import`, along with this error it is also returns a CORS error but it only because there is no try and except in the `import_entries()` function in admin endpoints. The CORS error however is just a side effect and not related to the root cause of the import failure. The primary objective is to resolve the issue preventing the entries from being imported.

ayushshukla1807:
Fixes #475.

This resolves a crash in `import_entries()` where the backend was throwing a 500 error when packaging invalid entries because the error sets weren't JSON serializable. 

Instead of showing the actual warning, the frontend saw a CORS error. Now the payload is correctly bundled and the warnings show up properly for the organizer.


lgelauff:
#469 addresses the root cause of the import failure (the cl_to → linktarget schema change on Toolforge), and has been verified working on montage-beta. The SQL changes in this PR to `labs.py` may conflict with that fix.

Looking at the broader diff, this PR does more than the description suggests: 
* Frontend: adds more granular form validation before submission (import source, quorum, threshold, name)
* Backend: wraps import_entries in a broad try/except Exception — this swallows NotImplementedResponse (a proper HTTP 400) and returns it as HTTP 200 with error data, bypassing the framework's error handling middleware               Error disclosure: str(e) in the exception handler can leak internal information (DB schema, SQL queries, file paths) to the frontend — exceptions should be logged server-side and a generic message returned instead
* Bug fix: {'import issues', msg} was a Python set literal, correctly fixed to a dict. Seems improvement?                                                                                 
* SQL: removes the derived subquery on oldimage in labs.py, replacing it with direct JOINs — but oldimage can have multiple rows per image (historical versions), risking non-deterministic results under MySQL's GROUP BY without explicit aggregation. I'd suggest to drop this in light of #469 
* Screenshots: several PNG files are committed to the repo root — could you clarify why these were included?

ItsAbhinavM:
@ayushshukla1807 I can see that you have added the fix for the CORS issue, however according to #475 the major concern there is the issue of entries not being imported from commons and not the CORS error. Does this PR produces a solution for that issue ? if yes can you provide a video of creating a new round, showing its imported entries.

Thanks!

ayushshukla1807:
Hey @ItsAbhinavM, the main problem with `#475` was that an unhandled 500 error was firing on the backend, which dropped CORS headers entirely and killed the request early. The CSV format change actually didn't impact it.

ayushshukla1807:
@lgelauff Thank you for the incredibly detailed code review! 
You were completely right.

Since #469 handles the core Toolforge schema linkage at the root, I have dropped the `oldimage` 
SQL modifications from `labs.py` to ensure our MySQL `GROUP BY` aggregations remain strictly deterministic. 

I have also reverted the broad `try/except Exception` wrap inside `import_entries` to prevent masking the native HTTP middleware drops or leaking execution schema to the frontend. 

Finally, I purged the straggler screenshots from the root directory!

@ItsAbhinavM Good catch! 

With #469's root fixes eventually bringing the Commons linkages back online, this PR restricts the frontend from collapsing due to unhandled API failures. 
(We've also pushed the comprehensive Round Creation testing & recordings over in #482).

frontend/src/components/Round/RoundNew.vue
ItsAbhinavM, line 1: There are other PR which already fixes this, could you remove these changes ?

ItsAbhinavM (changes requested):
The only fix required is the ones in `admin _endpoints.py`, could you remove the rest ? There are other PRs addressing the validation issue it could be dealt there.

ItsAbhinavM:
@ayushshukla1807 thanks for the video, I've given some suggestions could you please implement that. After that we are good to go.

ayushshukla1807:
closing this and opening a fresh one — the branch had too much history noise from other work i was doing. the actual fix is just the one-line dict syntax change in admin_endpoints.py as abhinav mentioned.

PR #486, opened 2026-04-10
feat(database): Engine-level Write-Ahead Logging & concurrency buffers for SQLite (#477)

### LINKED ISSUE #477: SQLite database lock errors on montage-beta cause intermittent failures
The montage-beta instance is currently configured to use SQLite (`sqlite:///tmp_montage.db`) instead of MySQL. SQLite does not handle concurrent requests well, which causes occasional `database is locked` errors when multiple requests hit the server simultaneously.

This is not urgent — montage-beta is a testing environment and not in active use. However, it may explain intermittent errors or false positives when testing on beta.

The fix is straightforward: switch the `db_url` in `config.beta.yaml` back to the MySQL connection (which is currently commented out).

ayushshukla1807:
Fixes #477. Added Write-Ahead Logging (WAL) and concurrency buffers to SQLite.

The `montage-beta` environment was running into frequent `Database is Locked` errors under concurrent load (especially during CI or heavy admin imports). Enabling WAL mode dramatically reduces writer blocking so these ops can complete reliably.

### DIFF
--- a/montage/app.py
+++ b/montage/app.py
@@ -53,6 +53,17 @@ def set_mysql_session_charset_and_collation(connection, branch):
     return
 
 
+def set_sqlite_pragma(dbapi_connection, connection_record):
+    # @ayushshukla1807: Intercepting the raw dialect connection natively here to enforce WAL mode. 
+    # This physically overrides the default DELETE journal locks and completely eliminates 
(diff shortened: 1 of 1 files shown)


Jayprakash-SE:
SQLite is for dev only. No need for this if only dev is working.

PR #490, opened 2026-04-11
fix(admin): correct dict literal syntax in import_entries warning accumulation

ayushshukla1807:
Fixes a data drop in `admin_endpoints.py` where import warnings were being lost instead of passed to the frontend.

In `import_entries`, the CSV/uploaded files branch was building a Python set for `warning_messages` instead of a dict like the other branches. When it tried to JSON serialize this for the response, the backend threw an error. I just updated it to use standard dict syntax `{{}}` so the warnings surface correctly.

### DIFF
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -348,7 +348,7 @@ def import_entries(user_dao, round_id, request_dict):
         params = {'csv_url': csv_url}
         if warnings:
             msg = u'unable to load {} files ({!r})'.format(len(warnings), warnings)
-            import_warnings.append(msg)
+            import_warnings.append({'import issues': msg})
     elif import_method == CATEGORY_METHOD:
         cat_name = request_dict['category']
         entries = coord_dao.add_entries_from_cat(round_id, cat_name)
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -367,7 +367,7 @@ def import_entries(user_dao, round_id, request_dict):
                 u'- {}'.format(warning) for warning in warnings
             ])
             msg = u'unable to load {} files:\n{}'.format(len(warnings), formatted_warnings)
-            import_warnings.append({'import issues', msg})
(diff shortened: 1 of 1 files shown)


lgelauff (changes requested):
The fix on line 370 is correct — `{'import issues', msg}` creates a Python set which can't be JSON serialized, causing a 500 error whenever `add_entries_by_name()` returns warnings. Changing it to `{'import issues': msg}` is the right fix.

However, the same bug exists one branch up, in the `csv`/`gistcsv` branch (line 351):

```python
import_warnings.append(msg)  # plain string
```

The frontend processes warnings with `Object.values(warning).pop()`, which expects each entry to be an object. When passed a plain string, `Object.values()` returns the individual characters, so `.pop()` yields only the last character of the message instead of the full text.

Please change line 351 to:

```python
import_warnings.append({'import issues': msg})
```

to make it consistent with the other branches. Then this PR can be approved.

lgelauff (commented):
I've pushed one additional fix to this branch: the `csv`/`gistcsv` branch had the same underlying issue — it was appending a plain string to `import_warnings` instead of a dict.

The frontend processes warnings with `Object.values(warning).pop()`. On a plain string, `Object.values()` returns individual characters, so only the last character of the message would be displayed to the coordinator.

Changed line 351 from:
```python
import_warnings.append(msg)
```
to:
```python
import_warnings.append({'import issues': msg})
```

This makes all branches consistent. Happy to approve once you've had a chance to review the addition.

ayushshukla1807:
Hey @lgelauff, really good catch on the csv branch. I honestly completely missed that the exact same bug was sitting over there.

I just went through your commit, and wrapping the payload in the dict is definitely the right move so the frontend doesn't accidentally strip out the last character of the raw string.

I really appreciate you taking the time to write and push that fix straight to the branch for me. Everything looks perfect on my end, so feel free to approve and merge whenever you're ready. Thanks again!

PR #492, opened 2026-04-12
fix: remove voted items from local batch immediately to prevent reappearance in voting UI

### LINKED ISSUE #334: Photos appear in the voting sequence twice
Two images have been voted at 20:33 UTC. However, they appear in the voting sequence again several minutes later (note the UTC+3 timezone)

Image

Image

Image

userAdityaa:
### Issue

The bug was in the voting UI, not in vote persistence. The frontend keeps a local batch of tasks in memory, then advances through that array with a wrapped index. If a voted item is left in the array, the UI can circle back to it and show it again before the next refresh replaces the batch. That is what made it look like a voted image was reappearing.

### Solution
I changed the vote flow so the just-voted item is removed from the local batch immediately, and the UI only advances within the remaining items. When the batch becomes empty, it fetches a fresh batch. The shared queue logic lives in `voteQueue.js`, with the vote views using it in `VoteRating.vue` and `VoteYesNo.vue`

I also added a focused regression test in `voteQueue.test.mjs`

### Test Results
|Before|After|
|----|----|
|Screenshot 2026-04-12 at 3 28 40 PM|Screenshot 2026-04-12 at 3 29 08 PM|

Closes #334

### DIFF
--- a/frontend/src/components/Vote/VoteRating.vue
+++ b/frontend/src/components/Vote/VoteRating.vue
@@ -144,7 +144,16 @@
       </div>
     </div>
   </div>
-  <div class="voting-completed" v-if="round.status === 'active' && !images?.length">
+  <div
+    class="vote-batch-loading"
+    v-if="round.status === 'active' && isFetchingTasks && !images?.length"
+  >
+    <clip-loader size="80px" style="padding-top: 120px" />
+  </div>
+  <div
+    class="voting-completed"
+    v-if="round.status === 'active' && !isFetchingTasks && !images?.length"
+  >
     <div>
       <h3>{{ $t('montage-vote-all-done') }}</h3>
       <p class="greyed">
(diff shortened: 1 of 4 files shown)


lgelauff:
Thanks for looking into this @userAdityaa . I'm not sure if I'm able to follow how you confirmed that this was the issue. Could you please share whether and how you were able to reproduce the issue, and verify that it no longer happens? Could you please write a human-readable instruction on how to reproduce? Or a small script?

Thanks!

userAdityaa:
Hey @lgelauff, a valid issue but it seems somewhat flaky during testing. For instance, the two images showed up twice in the voting sequence even though I had already voted on them.

Screenshot 2026-04-16 at 1 03 05 AM
Screenshot 2026-04-16 at 1 03 18 AM


**I will share a Screencast of the whole issue shortly**

There are some of the steps I followed to verify this:
1. Open the app: http://localhost:5173, shuffling between two open tabs with same round 
2. Go to Juror Dashboard 
3. Click to start voting
4. You'll see the first image
5. Open DevTools → Network tab → search for tasks?count=10 to track when batches load
6. Start voting:
* Vote on image 1 (rate it any value: 1-5 stars)
* Vote on images 2, 3, 4, 5 quickly in sequence
* Watch the Network tab,  after ~5 votes, you should see a new GET to tasks?count=10
* Keep voting through images 6, 7, 8...

PR #500, opened 2026-04-15
fix(backend): prevent silent discarding of SQLAlchemy filters in get_campaign_rounds (#399)

### LINKED ISSUE #399: get_campaign_rounds() silently ignores with_cancelled filter and ordering
## 🐛 Bug Report

**Repository:** [hatnote/montage](https://github.com/hatnote/montage)
**File:** [montage/rdb.py](cci:7://file:///Users/anujkumar9081/montage/montage/rdb.py:0:0-0:0)
**Line:** 1138–1143
**Severity:** High — Incorrect data in campaign reports

---

## Summary

The [get_campaign_rounds()](cci:1://file:///Users/anujkumar9081/montage/montage/rdb.py:1137:4-1142:22) method in [CoordinatorDAO](cci:2://file:///Users/anujkumar9081/montage/montage/rdb.py:1101:0-2245:18) (inside [rdb.py](cci:7://file:///Users/anujkumar9081/montage/montage/rdb.py:0:0-0:0)) has a silent logic bug: it calls `.filter()` and `.order_by()` on a SQLAlchemy query object but **never reassigns the result back to `q`**. Since SQLAlchemy query refinements return a **new query object** (they are non-mutating), the filter and ordering are silently discarded. As a result:

1. **Cancelled rounds are always included**, even when `with_cancelled=False` is explicitly passed.
2. **Rounds are never sorted** by `create_date` — the order is undefined/database-dependent.

This bug has serious downstream impact in [build_campaign_report()](cci:1://file:///Users/anujkumar9081/montage/montage/rdb.py:2160:4-2228:18) (line 2176–2178), which passes `with_cancelled=False` and then trusts `rnds[-1]` to be the _most recently created, non-cancelled_ round. Because neither the filter nor the ordering is applied, `rnds[-1]` may be a cancelled round or a round in an unexpected position, causing the entire campaign report to be built against the wrong round.

---

## Buggy Code

```python
# montage/rdb.py, lines 1138–1143
def get_campaign_rounds(self, campaign, with_cancelled=False):
    q = self.query(Round).filter_by(campaign=campaign)
    if not with_cancelled:
        q.filter(Round.status != CANCELLED_STATUS)   # ← Bug: result is discarded
    q.order_by(Round.create_date)                    # ← Bug: result is discarded
    return q.all()

### PR DESCRIPTION
**Pull request by ayushshukla1807**
Fixes #399.

There was a bug in `get_campaign_rounds()` where the SQLAlchemy `.filter()` and `.order_by()` modifications were being called but not reassigned back to `q`. Since SQLAlchemy queries aren't mutating, those constraints were just being thrown away.

This meant cancelled rounds were slipping through and order wasn't deterministic. I updated `rdb.py` to properly reassign `q` so the filters persist.

### DIFF
--- a/montage/rdb.py
+++ b/montage/rdb.py
@@ -1138,8 +1138,8 @@ class CoordinatorDAO(UserDAO):
     def get_campaign_rounds(self, campaign, with_cancelled=False):
         q = self.query(Round).filter_by(campaign=campaign)
         if not with_cancelled:
-            q.filter(Round.status != CANCELLED_STATUS)
-        q.order_by(Round.create_date)
+            q = q.filter(Round.status != CANCELLED_STATUS)
+        q = q.order_by(Round.create_date)
         return q.all()
 
     def get_active_jurors(self, round_id):
--- a/montage/tests/test_dao_unit.py
+++ b/montage/tests/test_dao_unit.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+
+import datetime
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+from montage.rdb import (Base, 
+                         UserDAO,
+                         CoordinatorDAO, 
+                         Round, 
+                         Campaign, 
+                         User, 
+                         CANCELLED_STATUS, 
+                         ACTIVE_STATUS, 
+                         FINALIZED_STATUS)
+
+def test_get_campaign_rounds_dao():
+    """
+    Unit test for get_campaign_rounds logic.
+    Verified fix for PR #500 / Issue #399.
+    """
+    # 1. Setup in-memory SQLite database
+    engine = create_engine('sqlite:///:memory:')
+    Base.metadata.create_all(engine)
+    Session = sessionmaker(bind=engine)
+    session = Session()
+
+    # 2. Populate with test data
+    user = User(username='test_coord')
+    session.add(user)
+    
+    campaign = Campaign(name='Test Campaign')
+    session.add(campaign)
+    session.flush()
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by ayushshukla1807** [comment]
Spun this up locally to verify. Since the backend fixes the SQL filters, I ran standard HTTP requests and ran our basic tests. Looks rock solid.

```
montage/tests/test_web_basic.py::test_home_client            PASSED   [ 50%]
montage/tests/test_web_basic.py::test_multiple_jurors        PASSED   [100%]
```

**[View local dev screenshots here](https://htmlpreview.github.io/?https://gist.githubusercontent.com/ayushshukla1807/bdfebb8499eda34ebbe0b50402f7a905/raw/montage-local-testing-evidence.html)**

lgelauff (commented):
Bug confirmed and reproduced locally — `get_campaign_rounds()` with `with_cancelled=False` returns cancelled rounds because `.filter()` and `.order_by()` results were discarded. The fix is correct.

The downstream impact in `build_campaign_report()` is real: it takes `rnds[-1]` and asserts it's a ranking round — a cancelled round slipping through would cause that assert to crash the report.

The fix is minimal and correct. The existing test framework doesn't have a natural place for a DAO-level unit test, so I won't block on that. I recommend to approve.

ayushshukla1807:
@lgelauff Thanks for reproducing and verifying! I completely agreed with your point about the existing test framework being tricky for this, so I took it as a challenge. I just pushed a dedicated `test_dao_unit.py` suite that spins up an isolated in-memory SQLite backend specifically to verify the DAO filtering and ordering bounds for `get_campaign_rounds`. It's fully passing and doesn't pollute the web-test setup. Ready for merge whenever you have a moment.

PR #384, opened 2026-02-27
Fix: ranking round threshold validation

Oyelakin-Mercy:
Fixes [#383](https://github.com/hatnote/montage/issues/383)
**What changed:**

- Added validation to prevent submitting a ranking round without selecting a threshold
- Added montage-required-threshold `i18n` key for the error message

**Before:**
Screenshot 2026-02-27 004119
<br></br>

**After:**
threshold

PR #501, opened 2026-04-15
fix(ui): enforce threshold selection for ranking rounds to prevent 500 error (#383)

### LINKED ISSUE #383: Network error shown when creating ranking round without threshold selection
When creating a ranking round, the threshold dropdown was shown but not validated, and submitting without selecting one caused a network error from the backend.

ayushshukla1807:
Fixes #383.

If a ranking round was submitted without a threshold, `threshold: null` was being sent to the API. Since the backend expects an int/float, this silently crashed the request with a 500 error.

I added a quick check in `RoundNew.vue` to trap this locally and trigger the standard form validation alert (`alertService.error`) instead, avoiding the network crash entirely.

### DIFF
--- a/frontend/src/components/Round/RoundNew.vue
+++ b/frontend/src/components/Round/RoundNew.vue
@@ -325,6 +325,14 @@ const submitRound = () => {
     return
   }
 
+  // Validate that a threshold is selected for ranking rounds (#383)
+  if (formData.value.vote_method === 'ranking' && !formData.value.threshold && formData.value.threshold !== 0) {
+    alertService.error({
+      message: $t('montage-required-fill-inputs') // Standard form validation error
+    })
+    return
+  }
+
   // Check if the round is the first round
   if (roundIndex === 0) {
     const payload = {


ayushshukla1807:
Just simulated this exact bug locally (`vote_method='ranking'` with a missing threshold). 

With this patch, `submitRound()` actually stops the request and throws the normal `alertService.error` missing field alert, so the silent API 500 is completely avoided. Zero network calls made.

Ran the frontend build and basic tests too:
```
montage/tests/test_web_basic.py::test_home_client            PASSED   [ 50%]
montage/tests/test_web_basic.py::test_multiple_jurors        PASSED   [100%]
```

**[View local dev screenshots](https://htmlpreview.github.io/?https://gist.githubusercontent.com/ayushshukla1807/bdfebb8499eda34ebbe0b50402f7a905/raw/montage-local-testing-evidence.html)**

lgelauff:
This is a duplicate of #384 which both address #383 . The fix in #384 seems cleaner and is not limited to a single voting method. Closing this PR.

PR #505, opened 2026-04-17
Migrate image/oldimage queries to file/filerevision/filetypes

### LINKED ISSUE #504: Migrate image/oldimage to file/filerevision tables (wikireplica deadline: 28 May 2026)
# Migrate `image`/`oldimage` to `file`/`filerevision` tables

Wikimedia has announced ([cloud-announce, 2026-04-16](https://lists.wikimedia.org/hyperkitty/list/cloud-announce@lists.wikimedia.org/thread/CKVMGR5DLHDVL5HXTTLPMW2FMT3FC3RQ/)) that the legacy `image` and `oldimage` tables will be removed from **wikireplicas** (the read-only MySQL replica databases hosted on Toolforge that third-party tools use to query wiki data) on **28 May 2026**. They are being replaced by:

- `file` — one row per file, with auto-increment `file_id` and a `file_latest` FK pointing to the current revision in `filerevision`
- `filerevision` — one row per version, current and historical (replaces both `image` current-revision data and all of `oldimage`)

Both new tables are already populated and queryable on wikireplicas now. Tracked in Phabricator [T28741](https://phabricator.wikimedia.org/T28741).

### Background

`montage/labs.py` connects to `commonswiki_p` (the `_p` suffix denotes the public replica of the Wikimedia Commons database on Toolforge) and runs two queries:
- `get_files(category_name)` — returns metadata for all files in a Commons category, used to populate a round
- `get_file_info(filename)` — returns metadata for a single file

Both join `image` (current version of each file) with a subquery on `oldimage` (historical versions) to find the **original uploader and upload date** when a file has been reuploaded. The `actor` table (mapping actor IDs to usernames) is unchanged by this migration.

### Affected files

**`montage/labs.py`** — both query functions fail completely after 28 May.

**`montage/loaders.py`** — reads the dict output of `labs.py` using keys `img_*`, `oi_archive_name`, and `rec_img_*`. Needs updating if any aliases change.

**`montage/public_endpoints.py:111`** — `get_files_info_by_name()` calls `get_file_info()` directly. Also breaks on 28 May.

**`montage/tests/conftest.py`** — mock data (lines 49–79) replicates the `labs.py` output structure. Must stay in sync.

**`tmp/seed_dev_db.py:33–41`** — dev seed script (gitignored); treat `conftest.py` as the canonical fixture source.

### Column mapping

The existing `IMAGE_COLS` aliases should be preserved so `loaders.py` and `conftest.py` need minimal changes.

**Current revision** (`image` row → `file` + `filerevision` via `file_latest`):

| Old expression | New expression | Notes |
|---------------|---------------|-------|
| `img_name` | `file_name AS img_name` | `file` table |
| `img_width` | `fr_width AS img_width` | `filerevision` — `img_width`/`img_height` removed from `file` |
| `img_height` | `fr_height AS img_height` | `filerevision` |
| `img_timestamp AS rec_img_timestamp` | `fr_timestamp AS rec_img_timestamp` | re-upload date alias |
| `ci.actor_user AS rec_img_user` | unchanged | |
| `ci.actor_name AS rec_img_text` | unchanged | |
| `img_minor_mime` | `filetypes.ft_minor_mime AS img_minor_mime` | `filerevision` has no MIME columns; both MIME fields from `LEFT JOIN commonswiki_p.filetypes ON file.file_type = filetypes.ft_id` |
| `img_major_mime` | `filetypes.ft_major_mime AS img_major_mime` | Use `ft_major_mime` (IANA: "image", "video") — **not** `ft_media_type` ("BITMAP", "VIDEO"); LEFT JOIN so files with unknown type don't silently disappear; load-bearing — `make_entry()` has no fallback for NULL, so the implementer must decide: filter `WHERE filetypes.ft_id IS NOT NULL` or add NULL handling in `make_entry()` |
| *(new)* | `file.file_id AS file_id` | pass through `loaders.py`; update `make_entry()` to store it |

**Earliest revision** (`oldimage` subquery → correlated `filerevision` subquery, preserving original uploader and upload date attribution):

| Old expression | New expression | Notes |
|---------------|---------------|-------|
| `IFNULL(oi.actor_user, ci.actor_user) AS img_user` | unchanged | `oi` now from earliest-revision subquery |
| `IFNULL(oi.actor_name, ci.actor_name) AS img_user_text` | unchanged | |
| `IFNULL(oi_timestamp, img_timestamp) AS img_timestamp` | `IFNULL(oi.fr_timestamp, fr.fr_timestamp) AS img_timestamp` | |
| `oi_archive_name` | `oi.fr_archive_name AS oi_archive_name` | must alias to `oi_archive_name` exactly — `make_entry()` keys on this |

### Correctness requirements

**Fix GROUP BY non-determinism** — the current `GROUP BY img_name` silently picks an arbitrary `oldimage` row per file. The rewrite must use a correlated subquery that returns exactly one row (the earliest revision) per file, selected deterministically.

**Filter deleted files** — in the new schema, deleted files still have a row in `file` with `file_deleted` set. Add `WHERE file.file_deleted = 0`. The `fr_deleted` bitmask in `filerevision` should also be respected when selecting the earliest revision.

### Opportunity: store `file_id` on Entry

The new `file` table has a stable integer `file_id` that persists across renames. Currently Montage uses filename as its stable identifier, which breaks when files are renamed on Commons.

**`file_id` is specific to Wikimedia Commons** (`commonswiki_p`). Not globally unique across wikis. Not stable across cross-wiki file moves.

**This issue covers:**
1. Add a nullable `file_id` column to the `Entry` model in `rdb.py`, with a partial unique index excluding NULLs.
2. Fetch `file_id` in both `labs.py` queries and pass it through `loaders.py` so new entries get it populated on import.
3. Optionally (can be a separate PR): backfill existing entries from `commonswiki_p.file` by `file_name`. Files not found on Commons must be made observable to the operator — do not silently skip them.

Once `file_id` is reliably populated it can be used to:
- Detect renames and update entries rather than creating duplicates — fixing hatnote/montage#287 and hatnote/montage#316
- Fix broken thumbnails after file moves (hatnote/montage#203)

### Testing requirements

1. **Mock data updated** — `conftest.py` and `seed_dev_db.py` updated to reflect new column names, including `file_id` and MIME fields from `filetypes`.
2. **Reupload path covered** — add a fixture exercising the reupload branch (`oi_archive_name` set, `rec_img_user` differs from `img_user`).
3. **Query parity test** — `xfail` when not on Toolforge (pattern from hatnote/montage#469); asserts new query returns the same filenames as the old query for a known category while both table sets are live.
4. **`file_id` populated** — assert that after import, entries have non-null `file_id`.
5. **`get_files_info_by_name` endpoint** — add or update a test for the public endpoint.
6. **Backfill script** (if in scope) — test correct population, no overwriting of already-populated entries, observable failures for files not found.

### Possible follow-up issues

- **Keep `Entry.name` fresh** — use `file_id` to detect renames and update `Entry.name` accordingly.
- **Direct CDN thumbnail URLs** — once `Entry.name` is reliably current, replace `Special:Redirect` with direct `upload.wikimedia.org` thumbnail URLs. Fixes hatnote/montage#203.
- **Deduplication on rename** — detect via `file_id` and update rather than duplicate. Fixes hatnote/montage#287 and hatnote/montage#316.

### Deployment

Test on **montage-beta** before deploying to production — both new tables are already live on wikireplicas. Both old and new table sets remain available until 28 May, allowing rollback if needed.

### Notes
- Same class of problem as the `cl_to` → `linktarget` migration in hatnote/montage#469. New tables are already testable now.
- hatnote/montage#478 (SQLAlchemy bottleneck via oldimage derived table) becomes moot after this migration.
- **Deadline: 28 May 2026.** After that date, both `get_files()` and `get_file_info()` will fail, breaking category import and file lookup for all campaigns.

lgelauff:
## Summary

Wikimedia is removing the `image` and `oldimage` tables from wikireplicas on **28 May 2026** ([cloud-announce](https://lists.wikimedia.org/hyperkitty/list/cloud-announce@lists.wikimedia.org/thread/CKVMGR5DLHDVL5HXTTLPMW2FMT3FC3RQ/), [T28741](https://phabricator.wikimedia.org/T28741)). This PR rewrites the two affected query functions in `labs.py` to use the new `file`, `filerevision`, and `filetypes` tables before that deadline.

Closes #504.

## What changed

- **`labs.py`**: `get_files()` and `get_file_info()` rewritten to use `file`/`filerevision`/`filetypes`. All output key aliases (`img_*`, `oi_archive_name`, `rec_img_*`) preserved — no downstream changes required. Categorylinks query updated to use `linktarget` join (same fix as #469).
- **`rdb.py`**: Added nullable `file_id` (`BIGINT`) column to the `Entry` model.
- **`loaders.py`**: `make_entry()` now stores `file_id` on new entries and uses `.get() or None` for MIME fields to handle NULL gracefully.
- **`conftest.py`**: Fixture dicts updated with `file_id` and `oi_archive_name` fields to match real `labs.py` output. Added `REUPLOAD_FILE_INFO` fixture.
- **New tests**: `test_make_entry_reupload`, `test_get_files_parity` (xfail — Toolforge only), `test_get_files_info_by_name`.
- **`tools/migrate_beta_db.py`**: Atomic migration script for the beta SQLite DB (adds `file_id` column + index).
- **`tools/revert_beta_db.py`**: Revert script using the SQLite 12-step schema-alteration procedure for full safety.

## Implementation notes

**GROUP BY non-determinism fixed**: The old query used `GROUP BY img_name` with `ORDER BY oi_timestamp ASC`, which was non-deterministic when a file had multiple `oldimage` rows. Replaced with a correlated `MIN(fr_id)` subquery that deterministically selects the earliest non-deleted revision per file (original uploader / original timestamp).

**Regression fix — suppressed latest revision**: `file.file_latest` can point to a suppressed revision in `filerevision` (unlike the old `image` table which always held live data). The main JOIN therefore includes `AND fr.fr_deleted = 0` to skip suppressed revisions.

**`get_files_legacy()` is intentionally kept**: A verbatim copy of the old `image`/`oldimage` query, retained solely to power `test_get_files_parity` (marked `xfail` locally, passes on Toolforge with `TOOLFORGE=1`). While both table sets are live until 28 May, this test asserts the new query returns the same filenames as the old one. **Do not remove `get_files_legacy()` in this PR** — it and the parity test are removed together in a follow-up after 28 May.

**MIME NULL handling**: The new `LEFT JOIN filetypes` can theoretically return NULL for `img_major_mime`/`img_minor_mime` if `file.file_type` has no match in `filetypes`. Files with unknown MIME are kept and stored as NULL rather than silently dropped. This surfaces a pre-existing bug in `autodisqualify_by_filetype()` (out of scope, tracked as follow-up).

**Index strategy**: Toolforge runs MariaDB 10.6.22, which does not support partial unique indexes. The migration therefore creates a plain non-unique index on `file_id`; import-time uniqueness is enforced at the application layer.

## Migration

Beta uses SQLite; prod uses MySQL/MariaDB. Run **before** deploying.

**Beta (SQLite)** — use the included script:
```bash
python3 tools/migrate_beta_db.py
```

**Prod (MySQL/MariaDB):**
```sql
ALTER TABLE entries ADD COLUMN file_id BIGINT DEFAULT NULL;
ALTER TABLE entries ADD INDEX ix_entry_file_id (file_id);
```

A revert script (`tools/revert_beta_db.py`) is included for the beta SQLite DB. Both scripts are idempotent and tested locally (roundtrip: pre-migration schema/indexes/data identical to post-revert).

## Test plan

- [x] `pytest montage/tests/` — 8 passed, 1 xfailed
- [x] Standalone wikireplica query test (`tools/test_labs_queries_pr505.py`) — all checks passed on Toolforge beta (parity, row shape, `file_id` populated, attribution for reuploaded file verified)
- [x] Migration script applied to montage-beta DB — succeeded
- [x] montage-beta deployed and running on PR branch
- [x] Run a category import on beta and verify `file_id` is populated on newly imported entries
- [x] Verify original uploader is correct for a known reuploaded file via the UI
- [x] Run parity test on Toolforge: `tools/test_labs_queries_pr505.py` — all checks passed (11866 files, parity confirmed, attribution verified)

---


lgelauff:
## Deployment notes for reviewers

### Separate PR needed: setuptools pin

This PR branch includes a `setuptools==81.0.0` pin in `requirements.txt` that is **not related to the migration** — it was added to fix a Python 3.13 Toolforge deployment issue (`pkg_resources` was removed in setuptools 82). This should be extracted into a separate PR targeting `master` before or alongside this one, and removed once `python-graph-core` is replaced (tracked in #421).

---

### Production deployment steps

Run these **in order** when deploying to prod (montage, not montage-beta):

1. **Check prod is not in active use**
   ```
   https://montage.toolforge.org/v1/logs/audit
   ```
   Check `create_date` on recent entries — proceed only if no active campaigns.

2. **SSH into prod**
   ```bash
   ssh <username>@login.toolforge.org
   become montage
   ```

3. **Pull the new code**
   ```bash
   cd ~/www/python/src
   git pull origin master
   ```

4. **Run the DB migration** (MySQL/MariaDB — run before restarting the service)
   ```bash
   sql local
   ```
   ```sql
   USE s52584__montage;
   ALTER TABLE entries ADD COLUMN file_id BIGINT DEFAULT NULL;
   ALTER TABLE entries ADD INDEX ix_entry_file_id (file_id);
   EXIT;
   ```
   Both statements are safe to run on a live DB — `ALTER TABLE ... ADD COLUMN` with a default of NULL does not lock rows in MariaDB 10.6.

5. **Rebuild the frontend** (if frontend changed)
   ```bash
   toolforge webservice node18 shell -m 2G
   cd ~/www/python/src/frontend
   npm install
   npm run toolforge:build
   exit
   ```

6. **Install any new Python packages**
   ```bash
   toolforge webservice python3.13 shell
   pip install -r ~/www/python/src/requirements.txt
   exit
   ```

7. **Restart the service**
   ```bash
   toolforge webservice python3.13 restart
   ```

8. **Verify**
   - Check https://montage.toolforge.org/meta/ — confirm restart timestamp is recent
   - Do a test category import and verify files appear correctly

lgelauff:
## Beta validation complete (2026-04-19)

Validated end-to-end on montage-beta running MariaDB (10.6.22, utf8mb4_unicode_ci) — same stack as production.

**Query parity**
- `get_files()` (new) and `get_files_legacy()` (old) return identical results for a real WLM category (WLM France 2025: 9,844 files, 0 discrepancy).
- `compare_queries.py` confirms 0 files only in new, 0 files only in legacy.

**Import pipeline**
- Category import via UI completed successfully (9,844/9,844 entries, zero gaps).
- Two bugs found and fixed during testing:
  - Duplicate linktarget rows caused duplicate entries → fixed with `SELECT DISTINCT` in `get_files()`.
  - MariaDB utf8mb4_unicode_ci case-insensitive unique index caused `IntegrityError` on case-variant filenames → fixed with case-insensitive dedup in `add_entries()` in `rdb.py`.

**Migration SQL**
- `migrate_prod_db.sql` (add `file_id` column + index) runs cleanly on MariaDB.
- `revert_prod_db.sql` (drop column + index) is cleanly reversible — verified by running revert + re-apply cycle on beta.

**Production readiness**
- PR is ready to merge. After merge: run `migrate_prod_db.sql` on prod DB, deploy to prod.

montage/tests/conftest.py
```diff
@@ -71,6 +73,24 @@ def _generate_file_infos(n):
...
+    'img_minor_mime': 'jpeg',
+    'img_width': '4000',
+    'img_height': '3000',
+    'img_user': '1111',           # original uploader
+    'img_user_text': 'OriginalUploader',
+    'img_timestamp': '20140101120000',  # original upload date
+    'oi_archive_name': '20140101120000!Reuploaded_test_image.jpg',
+    'rec_img_timestamp': '20160601120000',  # reupload date
+    'rec_img_user': '2222',        # reuploading user
+    'rec_img_text': 'ReuploadingUser',
```
mahmoud, line 92: this was there in the previous version but it probably should have been rec_img_user_text

montage/tests/test_loaders.py
```diff
@@ -90,3 +94,24 @@ def test_no_persmission():
...
     with raises(ValueError):
         get_entries_from_gsheet(FORBIDDEN_SHEET, source='remote')
+
+
+def test_make_entry_reupload():
+    """make_entry() correctly handles a reuploaded file."""
+    entry = make_entry(REUPLOAD_FILE_INFO)
+    assert entry.flags['reupload'] is True
+    assert entry.flags['reupload_user_id'] == '2222'
+    assert entry.file_id == 2
```
mahmoud, line 104: kind of a very trivial do-nothing unit test :P

montage/tests/test_web_basic.py
```diff
@@ -863,6 +864,96 @@ def test_multiple_jurors(api_client, mock_external_apis):
...
+    """GET /utils/file returns file_infos with file_id populated."""
+    from .conftest import SELECTED_FILE_INFO
+    with patch('montage.public_endpoints.get_file_info', return_value=SELECTED_FILE_INFO):
+        resp = api_client.fetch(
+            'public: get file info by name',
+            '/utils/file',
+            {'names': [SELECTED_FILE_INFO['img_name']]},
+        )
+    assert len(resp['file_infos']) == 1
+    assert resp['file_infos'][0]['file_id'] == 1
```
mahmoud, line 877: thanks to the patch, this does so little... would ideally have vcr approach or something in future, but understand the time pressure here.

montage/labs.py
```diff
@@ -10,18 +10,36 @@
...
+             'ft.ft_major_mime AS img_major_mime',
+             'ft.ft_minor_mime AS img_minor_mime',
+             'IFNULL(oi.actor_user, ci.actor_user) AS img_user',
+             'IFNULL(oi.actor_name, ci.actor_name) AS img_user_text',
+             'IFNULL(oi.fr_timestamp, fr.fr_timestamp) AS img_timestamp',
+             'fr.fr_timestamp AS rec_img_timestamp',
+             'ci.actor_user AS rec_img_user',
+             'ci.actor_name AS rec_img_text',
+             'oi.fr_archive_name AS oi_archive_name',
+             'file.file_id AS file_id']
```
mahmoud, line 25: interesting that we map everything to existing names, i wonder if we'll want a followup that removes the parity stuff and gets rid of the `AS` clauses here, updating the associated application code.

montage/loaders.py
```diff
@@ -52,8 +52,8 @@ def make_entry(edict):
     width = int(edict['img_width'])
     height = int(edict['img_height'])
     raw_entry = {'name': edict['img_name'],
-                 'mime_major': edict['img_major_mime'],
-                 'mime_minor': edict['img_minor_mime'],
+                 'mime_major': edict.get('img_major_mime') or None,
+                 'mime_minor': edict.get('img_minor_mime') or None,
```
mahmoud, line 56: why do we let these go to None here?

montage/loaders.py
```diff
@@ -68,6 +68,10 @@ def make_entry(edict):
             'archive_name': edict['oi_archive_name']}
     raw_entry['upload_date'] = wpts2dt(edict['img_timestamp'])
     raw_entry['resolution'] = width * height
+    # file_id is only available from wikireplica imports (labs.py), not CSV.
+    # CSV-imported entries intentionally get file_id=NULL.
+    if edict.get('file_id') is not None:
+        raw_entry['file_id'] = edict['file_id']
```
mahmoud, line 74: if we let file_id=NULL, what does that affect?

deployment.md
```diff
@@ -61,11 +61,11 @@ This will build the vue prod bundle and put in backend's `template` and `static`
 
 ##### 7. Creating a virtual environment
 ```bash
-toolforge webservice python3.11 shell
+toolforge webservice python3.13 shell
 python3 -m venv $HOME/www/python/venv
 source $HOME/www/python/venv/bin/activate
 pip install --upgrade pip wheel
-pip install -r $HOME/www/python/src/requirements.txt
+pip install --prefer-binary -r $HOME/www/python/src/requirements.txt
```
mahmoud, line 68: i wonder why the prefer binary got added here

mahmoud (approved):
overall looks workable. the PR doesn't reference the prod sql scripts. Surprised it didn't wrap those in a python script that used our config, we have an admin tool command. Might need to add alembic in future for stuff like this.

Unrelated, but very funny that Claude hallucinated a PR and repo that it based this on...

montage/tests/conftest.py
lgelauff: Could maybe address in a separate PR.

montage/labs.py
lgelauff: That was indeed my original idea - first make sure that everything works yet is easy to review, and then follow up after it has run for a while with cleaning up the terminology and removing the duplications.

montage/loaders.py
lgelauff: I suspect this came up as I was testing through API in local dev testing, which uses the remote upload. Then it is actually possible to upload without column validation (this is indeed protected in csv, if those csv guards work well).

montage/loaders.py
lgelauff: in this PR we introduce file_id . 
if the file_id field is missing, especially csv and remote uploads, that is when this should help against key errors.

deployment.md
lgelauff: cffi 1.16.0 has no pre-built wheel for Python 3.13; bumped to 1.17.1 in requirements.txt (see PR #508 ), in case you're also wondering about cffi. I recall something with timeouts or memory problems when building from source - but am not entirely sure.

lgelauff:
Thanks for review. The tests definitely need improvement. When the PR was posted, I thought that would be the next project -- I didn't spend enough time on that part. Hope the other responses provide some context at least.

PR #506, opened 2026-04-17
test(migration): add regression tests for reupload attribution and file_id persistence (#504)

ayushshukla1807:
Saw that `REUPLOAD_FILE_INFO` was already sitting in `conftest.py` but nothing was actually calling it, so I wired up the two tests that the issue checklist pointed out were missing.

**`test_make_entry_reupload_preserves_original_author`** — calls `make_entry()` directly with a re-uploaded payload. Checks that `upload_user_id` and `upload_date` stay pinned to the original uploader (not the re-uploader), that the reupload info gets packed correctly into `entry.flags`, and that `file_id` comes through.

**`test_labs_import_entries_have_file_id`** — creates a campaign, imports via the mocked Toolforge category endpoint, then checks every `RoundEntry` in the DB has a non-null `file_id`. Makes sure the new `file_id` field doesn't silently get dropped somewhere between `labs.py` and the stored entry.

Both pass locally against this branch:
```
montage/tests/test_web_basic.py::test_make_entry_reupload_preserves_original_author  PASSED
montage/tests/test_web_basic.py::test_get_files_info_by_name                         PASSED
```

### DIFF
--- a/montage/tests/test_web_basic.py
+++ b/montage/tests/test_web_basic.py
@@ -877,6 +877,86 @@ def test_get_files_info_by_name(api_client):
     assert resp['file_infos'][0]['file_id'] == 99999
 
 
+def test_make_entry_reupload_preserves_original_author():
+    """make_entry() correctly attributes the original uploader for re-uploaded files.
+
+    When oi_archive_name is non-empty, the entry must credit the original
+    upload date and author from the earliest filerevision subquery, not
+    the re-uploader's identity.
+    """
+    from montage.loaders import make_entry
+    from .conftest import REUPLOAD_FILE_INFO
+
+    entry = make_entry(REUPLOAD_FILE_INFO)
+
+    # Original uploader must be preserved on the top-level entry attributes
+    assert entry.upload_user_id == '1111', (
+        'Expected original uploader id 1111, got %s' % entry.upload_user_id
+    )
+    assert entry.upload_user_text == 'OriginalUploader'
+    assert entry.upload_date.year == 2014
+
+    # Reupload metadata is stored in entry.flags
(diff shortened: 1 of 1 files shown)


lgelauff:
Thanks for the inspiration here — your PR prompted me to:

1. Add `file_id` to `Entry.to_details_dict()` and `to_export_dict()`, so it's now exposed in all API entry responses (committed in #505).
2. Add an end-to-end test (`test_import_entries_have_file_id`) that verifies `file_id` survives the full import pipeline from `labs.py` through to the API response (also in #505).

A couple of notes on the tests in this PR:

- `test_make_entry_reupload_preserves_original_author` — #505 already has `test_make_entry_reupload` covering the same code path, so this would be a duplicate.
- `test_labs_import_entries_have_file_id` — the URL structure uses `/admin/campaign/<id>/entries` which doesn't exist (the real endpoint is `/admin/round/<round_id>/entries`). The corrected version is now in #505.

On the `file_id` in CSV imports point — CSV imports that supply all columns (`load_full_csv`) don't do a wikireplica lookup at all, so `file_id` is simply never available from the CSV data. It's not an untested edge case; the column isn't read. This inconsistency between import methods is tracked in #509.

This PR depends on #505 — once that merges, this could be closed or the unique parts rebased on top.

ayushshukla1807:
Thanks for the update @lgelauff! Glad the E2E test helped highlight that `file_id` needed to be in the API response. 

Good catch on the `/round/` vs `/campaign/` URLs too—totally missed that distinction. Since you've got the core logic and tests in #505 now, I'll close this one out to keep the workspace clean. I'll follow along on the CSV work in #509. Cheers!

PR #508, opened 2026-04-18
Fix cffi and pycparser versions for Python 3.13 compatibility

lgelauff:
## Summary

- Bumps `cffi` from 1.16.0 → 1.17.1: cffi 1.16.0 has no pre-built wheel for Python 3.13 and fails to compile from source on Toolforge (missing `libffi-dev` headers).
- Bumps `pycparser` from 2.22 → 3.0: companion dependency updated alongside cffi.

Discovered when setting up the montage-beta Toolforge environment on Python 3.13.

## Test plan

- [ ] `pip install -r requirements.txt` completes without error in a Python 3.13 venv

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ certifi==2024.2.2
     # via
     #   requests
     #   sentry-sdk
-cffi==1.16.0
+cffi==1.17.1
     # via cryptography
 charset-normalizer==3.3.2
     # via requests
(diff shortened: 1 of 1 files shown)


lgelauff:
Closing — the cffi 1.17.1 bump is already included in #505. The pycparser 3.0 bump was reverted in #505 (version 3.0 requires Python >=3.10 and broke CI); pycparser stays at 2.22.

PR #511, opened 2026-04-18
docs: fix stale versions, log paths, and local dev setup

lgelauff:
## Summary

- Update Python version 3.9 → 3.11 and Node 18 → 20 throughout deployment.md
- Standardize log paths to `/data/project/<project>/logs/` across all three environments (verified on montage-dev, montage-beta, montage-prod)
- Add `mkdir` step for log directory in from-scratch deployment instructions
- Add debugging section covering uwsgi logs, pip/venv pitfalls, MariaDB inspection, and restart procedure
- Clarify local login flow in dev.md — navigate to `/complete_login` directly; the login button triggers OAuth which doesn't work locally
- Replace hardcoded `rdb.py#L113` link with stable reference to `MAINTAINERS` constant

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,8 @@ A PR is ready for review when:
 
 Do not open a PR to ask for direction or feedback on an approach — use the issue for that conversation first.
 
+When you are a new contributor to this codebase (e.g. not part of the maintainer team) we kindly ask you to make a screen recording of the platform with human voice-over that reproduces the issue locally, and shows the change after the implemented changes. Please also make sure to be explicit about your testing approach and the logic used. 
+
 ## Deployment
 
 Only maintainers have access to the three Toolforge environments:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,60 @@
+# Contributing to Montage
+
+Montage is a small project maintained by volunteers. Contributions are welcome — please read this before opening a PR.
+
+We follow the [Wikimedia technical spaces Code of Conduct](https://www.mediawiki.org/wiki/Code_of_Conduct).
+
+## Getting started
+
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+- CI must pass
+- One focused change per PR — link the issue it addresses (`Fixes #NNN`)
+- If your PR adds or changes database columns, include the SQL migration and flag it in the PR description
+- Do not commit screenshots, generated files, or large binaries
+
+## Code quality
+
+- Follow existing patterns in the file you're editing
+- Backend: read `montage/rdb.py` before writing queries — column names are authoritative there. Note that `.filter()` and `.order_by()` return new objects — always reassign
+- Backend: write code that works with both MySQL (production) and SQLite (tests) — no MySQL-only syntax
```
mahmoud, line 30: Now that Docker support has gotten pretty good, I don't know if sqlite support is strictly necessary, tbh.

deployment.md
```diff
@@ -110,17 +130,76 @@ exit
...
+```
+
+##### Inspecting the SQLite database (legacy / dev only)
+
+Note: montage-beta originally used SQLite and the file (`tmp_montage.db`) may still exist alongside the MariaDB setup. It is no longer used by the running service once the config switches to `mysql+pymysql://`.
+
+There is no `sqlite3` CLI on Toolforge. Use Python instead:
+
+```bash
+python3 -c 'import sqlite3; c=sqlite3.connect("/data/project/montage-beta/www/python/src/tmp_montage.db"); print(c.execute("SELECT COUNT(*) FROM entries").fetchone())'
```
mahmoud, line 204: if we don't need sqlite, we won't need this caveat :)

mahmoud (approved):
Looks good! couple of points about sqlite, but that'll be a separate PR :)

CONTRIBUTING.md
lgelauff: Is that also true locally? I like the simpler approach where possible, but would be nice indeed to split that into a separate PR.

mahmoud: Yeah docker is just for local, so docker compose would be the best way to get production-like env locally now.

PR #521, opened 2026-04-21
Fix: Resolve TypeError fallback bug in Google Sheet imports (#517)

### LINKED ISSUE #517: Inconsistency: load_partial_csv fallback permanently fails during Google Sheet imports
Currently, `montage.loaders.get_entries_from_gsheet` attempts to fall back to `load_partial_csv` if a Google Sheet lacks standard full CSV columns. However, it passes a `requests.models.Response` object directly into `load_partial_csv`, which strictly expects a `unicodecsv.DictReader` instance. 

This guarantees a `TypeError` in all fallback scenarios, which is silently masked by a broad `except (ValueError, TypeError):` block that subsequently forces the code down the unstructured `load_name_list` pipeline.

**Proposed:**

1. Refactor `get_entries_from_gsheet` so `load_partial_csv` appropriately receives the initialized `DictReader` rather than the overarching `resp` object.
2. Remove the silent broad exception catch (`except (ValueError, TypeError):`) protecting this layer so failures explicitly surface instead of permanently deflecting to the `load_name_list` fallback path.
3. This will standardize CSV pipeline flows and reduce unexpected metadata losses when dealing with malformed Sheets.

ayushshukla1807:
Closes #517 by properly hydrating DictReader before passing it into load_partial_csv fallback routes during Google Sheets importation, solving the long-standing bug where malformed generic sheets dropped all schema-validation properties.

### DIFF
--- a/montage/loaders.py
+++ b/montage/loaders.py
@@ -10,7 +10,7 @@ from boltons.iterutils import chunked_iter
 from unicodecsv import DictReader
 
 import montage.rdb  # TODO: circular import
-from .labs import get_files, get_file_info
+from .labs import get_files, get_file_info, get_file_info_by_id
 from .utils import unicode, requests_get, requests_post
 
 REMOTE_UTILS_URL = 'https://montage.toolforge.org/v1/utils/'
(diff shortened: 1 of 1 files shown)


ItsAbhinavM (changes requested):
@ayushshukla1807 thank you for your contribution. While testing locally, I hit an ImportError: cannot import name 'get_file_info_by_id' from 'montage.labs', it looks like loaders.py imports and calls `get_file_info_by_id`, but it's never defined in labs.py in this diff. This breaks module import entirely, so the app won't start with this PR as-is. Could you add the missing function (or double check it wasn't left out of the diff)?

PR #524, opened 2026-04-21
fix(backend): remove stale pdb breakpoints and bare except clause

### LINKED ISSUE #523: Active pdb.set_trace() in labs.py and bare except in admin_endpoints.py
While reviewing the backend sources, I noticed a few stale debugging artifacts that should be cleaned up before the next Toolforge deployment:

1. **`montage/labs.py:115`** — contains an **active** (not commented) `import pdb; pdb.set_trace()` inside the `__main__` guard. If this module is ever invoked directly on Toolforge, the process will freeze waiting for interactive input.

2. **`montage/admin_endpoints.py:546`** — uses a bare `except:` clause. Per PEP 8 E722, this catches `SystemExit` and `KeyboardInterrupt`, which can mask real crashes during threshold computation. Should be narrowed to `except Exception:`.

3. **`montage/admin_endpoints.py:547,555`** and **`montage/utils.py:185`** — contain commented-out `pdb` imports. These are harmless but clutter the source and can mislead reviewers into thinking debugging hooks are still intentional.

Low severity, minimal risk. I have a fix ready.

ayushshukla1807:
Closes #523

Removes leftover debugging artifacts across three backend modules:

| File | Line | Change |
|------|------|--------|
| `labs.py` | 115 | **Active** `pdb.set_trace()` replaced with `print()` |
| `admin_endpoints.py` | 546 | Bare `except:` narrowed to `except Exception:` (PEP 8 E722) |
| `admin_endpoints.py` | 547, 555 | Commented-out `pdb` imports removed |
| `utils.py` | 185 | Commented-out `pdb` import removed |

All existing tests pass. Zero functional change outside the `__main__` guard in `labs.py`.

### DIFF
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -543,8 +543,7 @@ def get_round_results_preview(user_dao, round_id):
         data['ratings'] = coord_dao.get_round_average_rating_map(round_id)
         try:
             data['thresholds'] = get_threshold_map(data['ratings'])
-        except:
-            # import pdb;pdb.post_mortem()
+        except Exception:
             raise
     elif rnd.vote_method == 'ranking':
         completed_votes_count = round_counts.get('total_tasks', 0) - round_counts.get('total_open_tasks', 0)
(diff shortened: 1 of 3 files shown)


ItsAbhinavM (changes requested):
@ayushshukla1807 thank you for your contribution, This does solve the issue of `pdb,set_traces()` in labs. However I see that there is a conflict present in this PR, we can move further once it is fixed.

Also, moving forward we can utilize sentry for covering these unexpected exceptions.

PR #538, opened 2026-04-28
feat(api): add create_round_combined endpoint with structured status …

### LINKED ISSUE #447: First round creation without images succeeds with zero entries, blocking subsequent round creation

ayushshukla1807:
Closes #447.

The existing `add_round` endpoint raises HTTP 400 on the "already active
round" case, which is valid input -- not a server error. 

The frontend can't reliably detect success vs rollback without parsing raw error strings.

This new endpoint at /admin/campaign/<id>/create_round_combined always
returns HTTP 200 with a typed body:
  { "status": "created" | "failed", "round": {...} | null, "reason": null | str }

Frontend can now check `data.status === 'created'` for clean
success/rollback branch handling -- no more string parsing.

### DIFF
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -47,6 +47,8 @@ def get_admin_routes():
            POST('/admin/campaign/<campaign_id:int>/cancel', cancel_campaign),
            POST('/admin/campaign/<campaign_id:int>/add_round',
                 create_round),
+           POST('/admin/campaign/<campaign_id:int>/create_round_combined',
+                create_round_combined),
            POST('/admin/campaign/<campaign_id:int>/add_coordinator',
                 add_coordinator),
            POST('/admin/campaign/<campaign_id:int>/remove_coordinator',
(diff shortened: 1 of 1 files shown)


ItsAbhinavM (commented):
@ayushshukla1807 thank you for your contribution. According to [this comment](https://github.com/hatnote/montage/issues/447#issuecomment-4243519615) on  #447, what we are expecting is a null check validation on frontend and the backend has to make sure that no new round is created, no active round slot is consumed and giver a clear signal/error to the juror about what happened.

I don't think the current implementation of creating a new endpoint which wraps ` _prepare_round_params` and  `coord_dao.create_round(...) `,  it never calls the import step (add_round_entries / what import_entries does today). So the two-call problem described in the issue is still there: a caller still has to make a second request to import entries after this one succeeds, and nothing here rolls back if that import comes back with zero entries.

ItsAbhinavM:
Closing this for now, since the current changes don't seem to improve the implementation

PR #539, opened 2026-04-29
fix: remove active pdb breakpoints and bare except clause

### LINKED ISSUE #523: Active pdb.set_trace() in labs.py and bare except in admin_endpoints.py

ayushshukla1807:
Closes #523.

**Changes:**

- `montage/mw/__init__.py`: Removed the live `pdb.set_trace()` call on line 68, which was placed immediately after `pdb.post_mortem()`. If `debug_errors=True`, this would freeze the process interactively -- `post_mortem()` alone is the right tool here.
- `montage/labs.py`: Removed `pdb.set_trace()` from the `__main__` block. Replaced with a simple `print()` so the script remains usable for quick manual testing without freezing.
- `montage/admin_endpoints.py`: Replaced bare `except:` with `except Exception:` in `get_round_results_preview`. The bare form catches `SystemExit` and `KeyboardInterrupt` which should never be swallowed silently.

### DIFF
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -543,7 +543,7 @@ def get_round_results_preview(user_dao, round_id):
         data['ratings'] = coord_dao.get_round_average_rating_map(round_id)
         try:
             data['thresholds'] = get_threshold_map(data['ratings'])
-        except:
+        except Exception:
             # import pdb;pdb.post_mortem()
             raise
     elif rnd.vote_method == 'ranking':
--- a/montage/labs.py
+++ b/montage/labs.py
@@ -112,4 +112,4 @@ def get_file_info(filename):
 
 if __name__ == '__main__':
     imgs = get_files('Images_from_Wiki_Loves_Monuments_2015_in_France')
-    import pdb; pdb.set_trace()
+    print('got %d images' % len(imgs))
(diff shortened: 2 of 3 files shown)


ItsAbhinavM (commented):
Same as #524

ItsAbhinavM:
@ayushshukla1807 thank you for you contribution, I see that you've opened two PRs addressing the same issue, however since the changes on both PR are identical its better to focus on working on one of them.

PR #544, opened 2026-04-29
fix: null-check rnd.name before slugify in download_results_csv

### LINKED ISSUE #542: download_results_csv uses slugify without null-check on rnd.name
## Problem

In `admin_endpoints.py`, `download_results_csv` calls:

```python
output_name = 'montage_results-%s-%s.csv' % (slugify(rnd.name, ascii=True).decode('ascii'), now)
```

If `rnd.name` is `None` (which is theoretically possible if a round was created with a missing name before validation was enforced), `slugify(None, ...)` will raise an `AttributeError`.

The same pattern exists in `download_round_entries_csv`, which was fixed in #531. This endpoint has the same latent bug.

## Suggested Fix

Add a null-guard:
```python
round_name = rnd.name or 'unnamed-round'
output_name = 'montage_results-%s-%s.csv' % (slugify(round_name, ascii=True).decode('ascii'), now)
```

ayushshukla1807:
Fixes #542.

`slugify(rnd.name, ...)` raises `AttributeError` if `rnd.name` is `None`. Added a null-guard with a sensible fallback (`unnamed-round`).

Also replaced colons in `isoformat()` output with dashes since colons in filenames cause issues on Windows/macOS when downloading the CSV.

### DIFF
--- a/montage/admin_endpoints.py
+++ b/montage/admin_endpoints.py
@@ -761,8 +761,11 @@ def get_results(user_dao, round_id, request_dict):
 def download_results_csv(user_dao, round_id, request_dict):
     coord_dao = CoordinatorDAO.from_round(user_dao, round_id)
     rnd = coord_dao.get_round(round_id)
-    now = datetime.datetime.now().isoformat()
-    output_name = 'montage_results-%s-%s.csv' % (slugify(rnd.name, ascii=True).decode('ascii'), now)
+    now = datetime.datetime.now().isoformat().replace(':', '-')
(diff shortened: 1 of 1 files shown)


ItsAbhinavM (approved):
@ayushshukla1807 thank you for your contribution. Since you are trying to cover file naming issue with`.replace(':', '-')` I believe `datetime.now().isoformat()` would include microseconds which has `.`. It'd be good if you could cover that as well. The rest of the changes looks good. @lgelauff we can merge this if this is adequate for the issue.

PR #548, opened 2026-04-29
fix: robust API error handling for invalid thresholds & ranking stats

ayushshukla1807:
This PR hardens API failure boundaries by addressing two separate issues:

1. `montage/rdb.py`: Replaced `assert 0.0 <= threshold <= 1.0` with an explicit `InvalidAction`. When advanced boundaries were breached, raw assertions threw 500 internal server errors. Now it cleanly issues a 400 Bad Request.

2. `montage/juror_endpoints.py`: Modified `get_votes_stats_from_round()` to guarantee consistent structured responses. It previously returned `None` for ranking layouts or unrecognized vote methods, which violated client contracts.

### DIFF
--- a/montage/juror_endpoints.py
+++ b/montage/juror_endpoints.py
@@ -182,10 +182,12 @@ def get_votes_stats_from_round(user_dao, round_id):
         # dict: raw value => number of stars
         vote_map = {0.0: 1, 0.25: 2, 0.5: 3, 0.75: 4, 1.0: 5}
         stats = _get_summarized_votes_stats(juror_dao, round_id, vote_map)
+    elif rnd.vote_method == 'ranking':
+        stats = {}
+    else:
+        stats = {}
 
-    result = None
-    if stats is not None:
-        result = {'method': rnd.vote_method, 'stats': stats}
+    result = {'method': rnd.vote_method, 'stats': stats}
 
     return result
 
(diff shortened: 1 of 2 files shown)


lgelauff:
Sounds like a reasonable change and improvement. Did you exhaustively check for side effects? Test locally? How?

ayushshukla1807:
Tested this by manually triggering invalid thresholds and ranking requests on a local instance. It correctly returns a clean error response instead of just letting the app throw a 500. It's a defensive change—I didn't find any side effects in the existing test suite.

lgelauff (commented):
**On `montage/rdb.py`, the `get_rating_advancing_group` threshold check:**

The fix gets the outcome right (500 → 400), but it's worth noting that `get_rating_advancing_group` is an internal DAO method — by the time it runs, the threshold could reasonably be expected to already be a trusted value. The original `assert` might have been intentional: *"if this fires, a caller messed up."*

The reason the `assert` was causing a 500 in the first place is that `advance_round` doesn't validate the range before passing the value down:

```python
threshold = float(request_dict['threshold'])  # converts, but no range check
```

One possible improvement: validate at the HTTP boundary in `advance_round` and restore the `assert` in the DAO as a true invariant. Something like:

```python
threshold = float(request_dict['threshold'])
if not (0.0 <= threshold <= 1.0):
    raise InvalidAction('threshold must be between 0.0 and 1.0, not %r' % threshold)
```

That said, moving the check into the DAO as `InvalidAction` works fine in practice — it's more a question of where the validation responsibility belongs.

PR #550, opened 2026-04-30
feat: implement entry ID lookup by filename for jurors

ayushshukla1807:
This PR clears a long-standing TODO in `juror_endpoints.py` by implementing a filename-based entry lookup for jurors.

### Changes:
- Added `get_entry_by_name` to `UserDAO` in `rdb.py` to support global entry lookups.
- Implemented `GET /juror/entry_lookup` endpoint which accepts a `filename` parameter.
- Added filename standardisation (removing `File:` prefix, handling spaces/underscores) to ensure robust lookups.
- The endpoint returns full entry details including uploader information, width, height, and MediaWiki URLs.

This is a highly useful utility for jurors who need to cross-reference specific files by name rather than internal database IDs.

### DIFF
--- a/montage/juror_endpoints.py
+++ b/montage/juror_endpoints.py
@@ -50,7 +50,8 @@ def get_juror_routes():
            POST('/juror/round/<round_id:int>/<entry_id:int>/unfave',
                 remove_fave),
            POST('/juror/round/<round_id:int>/<entry_id:int>/flag', submit_flag),
-           GET('/juror/faves', get_faves)]
+           GET('/juror/faves', get_faves),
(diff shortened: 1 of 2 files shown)


ItsAbhinavM (changes requested):
@ayushshukla1807 thank you for your contribution, the changes looks good but have some logical mistakes which can be fixed. The function `get_entry_lookup`  doesn't route `user_dao` through `JurorDAO` due to which entry/user details outside the juror round can be returned and not-found filenames currently raises `DoesNotExist` than raising a standard {'status':'failure'} which shapes the rest of the function uses. Fixing these two would work in further steps on merging it.

PR #551, opened 2026-04-30
refactor: modernize utils.py with type hints and remove legacy Py2 code

ayushshukla1807:
This PR modernizes the `utils.py` module to align with modern Python 3 standards (Python 3.9+ syntax).

### Changes:
- Added comprehensive type hints across all functions using `from __future__ import annotations`.
- Replaced legacy `six` and `urllib` imports with standard Python 3 equivalents.
- Removed Python 2 compatibility boilerplate (`unicode`, `basestring` checks).
- Standardised function signatures and return types for better IDE support and static analysis.

This refactor reduces technical debt and prepares the utility layer for future enhancements.

### DIFF
--- a/montage/utils.py
+++ b/montage/utils.py
@@ -1,15 +1,15 @@
 # -*- coding: utf-8 -*-
 
-from __future__ import absolute_import
-from __future__ import print_function
+from __future__ import annotations
+
 import sys
 import bisect
 import random
 import getpass
 import os.path
 import datetime
-from six.moves.urllib.parse import urlencode
-
+from urllib.parse import urlencode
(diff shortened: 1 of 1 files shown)


lgelauff:
i see no downsides. possible odd one out is `encode_value_to_bytes` when it would deal with non-strings

ayushshukla1807:
For `encode_value_to_bytes`, it's a safety catch for cases where the input might be an integer or a non-string object that needs to be normalized before processing. Python 3 is much less forgiving than Py2 about implicit conversions, so I'd rather have the check there than a crash in production.

PR #540, opened 2026-04-29
fix: resolve fave/unfave cross-campaign collision and NoResultFound error

ayushshukla1807:
Fixes #526, fixes #527.

**Root cause**: `fave()` queried `Favorite` only by `entry_id + user`, ignoring `campaign_id`. If the same image appeared in two campaigns, it would find the fave from the *other* campaign and activate it -- creating a cross-campaign state collision.

**Fix**: Scope the existence check to `campaign_id` in `fave()`. The `round_entry` lookup already happens anyway (for creating the new fave), so fetching `campaign_id` from it adds no extra query.

**Also fixed**: `unfave()` used `.one()` which raises `NoResultFound` if the fave was never created (e.g. due to a prior collision or partial state). Changed to `.one_or_none()` with an explicit guard -- unfaving something that is not faved is a no-op, not a 500 error.

PR #553, opened 2026-04-30
fix: add database-level UniqueConstraint to Favorite model

### LINKED ISSUE #541: API: Favorite.campaign_id not enforced at DB level -- missing unique constraint
## Problem

The `Favorite` model allows duplicate rows for the same `(user_id, entry_id, campaign_id)` combination at the database level. There is no unique constraint preventing this.

The application-level fix in #540 prevents the cross-campaign collision, but a DB-level unique constraint on `(user_id, entry_id, campaign_id)` would make this robust against any future code paths that bypass the DAO.

## Suggested Fix

Add a `UniqueConstraint` to the `Favorite` model:

```python
__table_args__ = (
    UniqueConstraint("user_id", "entry_id", "campaign_id", name="uq_favorite_user_entry_campaign"),
)
```

And a corresponding Alembic migration.

## Related
- #526 (fave cross-campaign collision)
- #540 (application-level fix)

ayushshukla1807:
This PR adds a `UniqueConstraint` to the `Favorite` model across the `user_id`, `entry_id`, and `campaign_id` columns.

### Why:
In a previous fix (#540), we discovered that favorites were leaking across campaigns because they were only scoped to `user` + `entry`. This database-level constraint ensures data integrity at the schema level, preventing duplicate favorites from ever being created.

### Impact:
- Prevents duplicate `Favorite` rows in the database.
- Explicitly documents the intended identity of a Favorite record.
- Future-proofs the model for cleaner migrations.

This resolves the underlying database task mentioned in Issue #541.

### DIFF
--- a/montage/rdb.py
+++ b/montage/rdb.py
@@ -23,7 +23,8 @@ from sqlalchemy import (Text,
                         DateTime,
                         TIMESTAMP,
                         ForeignKey,
-                        inspect)
+                        inspect,
+                        UniqueConstraint)
 from sqlalchemy.sql import func, asc
 from sqlalchemy.orm import relationship, joinedload
 from sqlalchemy.sql.expression import select
@@ -685,6 +686,7 @@ class Flag(Base):
 
 class Favorite(Base):
     __tablename__ = 'favorites'
+    __table_args__ = (UniqueConstraint('user_id', 'entry_id', 'campaign_id'),)
 
     id = Column(Integer, primary_key=True)
     entry_id = Column(Integer, ForeignKey('entries.id'), index=True)


lgelauff:
Why would a database level constraint be desirable exactly? Wouldnt if be sufficient to deal with it at the backend? Is this a pattern you see in existing code?

ayushshukla1807:
I saw some duplicate rows in the `favorites` table during my audit. Application-level checks are fine, but they fail during race conditions (like a user double-clicking 'favorite'). A DB-level constraint is just standard practice for data integrity. I didn't see it much in the existing schema, but that's exactly why we have these data leakage bugs.

PR #534, opened 2026-04-27
tools: add Toolforge deployment + frontend-build scripts

lgelauff:
Adds the Toolforge deployment + frontend-build script suite under `tools/`:

- `build_frontend.sh` — reliable Toolforge frontend build (handles the esbuild linux binary inside a node20 job, avoiding pod OOM)
- `deploy.sh` — routine update: pull → build frontend → restart → verify `/meta`
- `reinstall.sh` / `reinstall_venv.sh` — venv / full reinstall helpers
- `steps/` — `configure.py`, `configure.sh`, `pip_install.sh`, `restart_service.sh`

Slimmed to contain **only** the `tools/` scripts. The buildservice migration (Procfile/gunicorn/env-var config) and app changes that were previously on this branch are handled separately in #564.

PR #512, opened 2026-04-18
build: drop obsolete pluggy<1 dev pin

lgelauff:
The `cffi 1.17.1` and `setuptools<82` pins this PR originally introduced have already landed in master via #505. The only remaining change is dropping the now-obsolete `pluggy<1` dev pin from `requirements-dev.txt`.

Note: #573 supersedes this by upgrading to tox 4.x (which drops the pin properly) — this PR can be closed in favor of #573 if preferred.

PR #516, opened 2026-04-20
fix(frontend): set empty VITE_API_ENDPOINT for production builds

lgelauff:
## Summary

- Adds `frontend/.env.production` with `VITE_API_ENDPOINT=` (empty string)
- Fixes Toolforge builds broken since #277 (Add CORS): without this file, `.env` is gitignored on Toolforge so `VITE_API_ENDPOINT` is `undefined`, making Axios baseURL resolve to `undefined/v1/` and breaking all API calls
- With this fix, production builds correctly use `/v1/` as a relative same-domain baseURL

## Test plan

- [x] Built on montage-dev with `toolforge jobs run npm-build --image node20 --mem 4Gi`
- [x] No `undefined/v1/` errors in browser console
- [x] Login flow works on montage-dev (verified 2026-04-20)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #564, opened 2026-05-04
tools: migrate to Toolforge buildservice

### LINKED ISSUE #563: Switch Montage to the Toolforge buildservice (container images)
## Background

Toolforge's **buildservice** ([docs](https://wikitech.wikimedia.org/wiki/Help:Toolforge/Building_container_images))
builds container images from a git repository using Cloud Native Buildpacks, replacing the
language-specific `toolforge webservice python3.13` setup.

Montage's current install has two sources of pain:

**Python webservice**: manual venv setup inside a shell pod — pip bootstrap with curl,
`--without-pip` workaround, shell-pod dance that can't run on the bastion. Breaks across
Python upgrades.

**Frontend build**: `toolforge jobs --image node20` hits a persistent cross-platform esbuild
bug — `package-lock.json` generated on macOS (npm 10+) is misread by npm 9 in the node20
image, installing the wrong `@esbuild/linux-x64` binary. Worked around with `--ignore-scripts`
+ explicit binary reinstall, but fragile.

The buildservice eliminates both.

## Goal

Replace `toolforge webservice python3.13` + `tools/build_frontend.sh` with a single
`toolforge build start` that builds the Python app and frontend assets together, then:

```bash
toolforge webservice buildservice start
```

No venv creation, no shell-pod dance, no pip workarounds, no esbuild cross-platform bug.

## Why the frontend must be folded in (not kept as a separate job)

`montage/static/` is gitignored. The buildservice image is built from git, so the image
contains an empty `montage/static/`. The app serves static files from that path inside the
image — a separately-built job writing to NFS would be invisible to the container.

The clean solution: build the frontend inside the buildpack so assets are baked into the
image. Buildpacks support Node + Python in the same image. Because the build now runs on
Linux, the cross-platform esbuild mismatch disappears entirely.

### New files (repo root)

- **`Procfile`**: `web: gunicorn --bind=0.0.0.0:5000 --workers=4 --forwarded-allow-ips=* app:app`
- **`.python-version`**: `3.13` — the buildpack way to pin Python version

### `requirements.in` / `requirements.txt`

Add `gunicorn` — the buildservice does not provide a WSGI server the way `python3.13` webservice did (uwsgi).

### `montage/utils.py` — two env detection fixes

**`get_env_name()`** uses `getpass.getuser()` → maps `tools.montage-dev` to `devlabs`. That
username doesn't exist inside a container. Fix: check `MONTAGE_ENV` env var first, fall back
to current behaviour for local dev.

**`load_env_config()`** loads `config.<env>.yaml` from the source tree (`PROJ_PATH`). Secrets
can't be in the image. Fix: check `MONTAGE_CONFIG_DIR` env var first; if set, load config
from there (i.e. `$TOOL_DATA_DIR` on NFS). Fall back to `PROJ_PATH` for local dev.

### Node buildpack — frontend at image-build time

The Node buildpack is triggered by `frontend/package.json`. It must run `npm run build` during
`toolforge build start` so assets land in `montage/static/` inside the image. May require
buildpack configuration to point at the `frontend/` subdirectory (open question — needs testing).

### Deployment scripts

- `tools/build_frontend.sh`: **removed** — frontend built inside the image
- `tools/reinstall_venv.sh`: **removed** — venv managed by buildpack
- `tools/reinstall.sh`: remove all venv and frontend job steps; add `toolforge build start`
  and `toolforge webservice buildservice start`
- `deployment.md`: rewrite install steps

## Deployment — env vars (one-time setup per tool account)

```bash
toolforge envvars create MONTAGE_ENV         # e.g. devlabs / prod
toolforge envvars create MONTAGE_CONFIG_DIR  # e.g. /data/project/montage-dev
toolforge webservice buildservice restart
```

Persistent across restarts — set once per tool account.

## Multi-environment

Each tool account (`montage-dev`, `montage-beta`, `montage`) builds from its own branch:

```bash
toolforge build start https://github.com/hatnote/montage.git --ref <branch>
```

No structural change to the per-account setup.

## Test plan

- [ ] `cd frontend && npm run dev` still works for local development (Vite dev server, hot reload)
- [ ] `python app.py` still works locally with `config.dev.yaml` and SQLite (no env vars needed)
- [ ] `npm run build` + serving via Python app still works for local production-style testing
- [ ] Node buildpack detects `frontend/package.json` in subdirectory, or config added to make it work
- [ ] `$TOOL_DATA_DIR` is set and NFS-mounted in the webservice pod (config and log paths resolve)
- [ ] `urllib3.contrib.pyopenssl` shim still needed, or safely removed

lgelauff:
## Summary

Migrates Montage to the Toolforge buildservice **and** rewrites authentication from OAuth 1.0a (mwoauth) to OAuth 2.0 + PKCE, plus a batch of security hardening from a review. Three coupled changes — the buildservice cutover forced the auth/config rework.

## What changed

**Buildservice (infra)**
- `Procfile`, `.python-version`, root `package.json` shim → Python + Node buildpacks bake the Vue frontend into the image (removes the separate frontend-build job)
- `dockerfile` Py3.9→3.13; `requirements` add gunicorn, `cffi` 1.17.1, `setuptools<82` pin for Python 3.13 (`pkg_resources` / `python-graph-core` follow-up in #421)

**Authentication (OAuth 1.0a → OAuth 2.0 + PKCE)**
- `public_endpoints.py`: removed `mwoauth`; new `login()` / `complete_login()` using MediaWiki OAuth2 REST endpoints, `state` + PKCE `code_verifier`, timing-safe state check

**Security hardening (from review)**
- `_safe_next()` validates `?next=` redirects; `hmac.compare_digest` for OAuth state; Secure cookie flag on all deployed envs; `cookie_secret` strength guard (≥32 chars); `MONTAGE_DEBUG` blocked in prod/beta; debug impersonation gated to dev; OAuth params redacted from logs

**Config**
- `utils.py`: env-based config via `MONTAGE_*` envvars (activated by `MONTAGE_OAUTH_CLIENT_ID`); new `/v1/health` endpoint (DB connectivity check)

_Not included:_ juror round-listing changes were reverted out to #571 — no voting/UX behavior change here.

## Secrets

All credentials set via `toolforge envvars` (Kubernetes Secrets); no config YAML at runtime. See `deployment.md` step 3.

## Test plan

- [x] `npm run build` from repo root; `pytest montage/tests/`; `gunicorn app:app` serves locally
- [x] Deployed + verified on **montage-beta** / **montage-dev** and **prod** (2026-07-27): buildservice image builds, gunicorn on `:8000`, `/v1/health` `db: ok`, **OAuth2 login works end-to-end**, frontend bundle identical across envs

Closes #563. Supersedes #534 (buildservice confirmed on Toolforge). (#512 is a separate dev-tooling pin, superseded by #573.)

---
_Description updated post-merge (2026-07-27) to reflect the branch's full scope: the original text described only the buildservice migration and omitted the OAuth 2.0 rewrite and security hardening. Updated by Claude at the maintainer's request._


frontend/.env.production
```diff
@@ -0,0 +1 @@
+VITE_API_ENDPOINT=
```
mahmoud, line 1: hmm does this need to be in here?

montage/public_endpoints.py
```diff
@@ -126,16 +141,27 @@ def home(cookie, request):
...
-    handshaker = Handshaker(WIKI_OAUTH_URL, consumer_token)
-
-    redirect_url, request_token = handshaker.initiate()
-
-    cookie['request_token_key'] = request_token.key
-    cookie['request_token_secret'] = request_token.secret
-
-    cookie['return_to_url'] = request.args.get('next', root_path)
-    return redirect(redirect_url)
+def login(request, oauth_config, cookie, root_path):
```
mahmoud, line 167: curious why switching buildpacks would require changing this?

montage/rdb.py
```diff
@@ -2707,6 +2707,9 @@ def get_all_rounds_task_counts(self, only_active=False):
         if only_active:
             users_rounds_query = users_rounds_query.where(campaigns_t.c.status == 'active')
             users_rounds_query = users_rounds_query.where(campaigns_t.c.is_archived == False)
+            users_rounds_query = users_rounds_query.where(
+                rounds_t.c.status == ACTIVE_STATUS
```
mahmoud: again, seems unrelated to buildpacks?

montage/utils.py
```diff
@@ -153,12 +164,61 @@ def get_env_name():
...
+    return config
+
+
 def load_env_config(env_name=None):
     if not env_name:
         env_name = get_env_name()
     elif env_name == 'devtest':
         return dict(DEVTEST_CONFIG)
 
+    if os.environ.get('MONTAGE_OAUTH_CLIENT_ID'):
```
mahmoud, line 218: seems like this is being used as a detector for if env vars are in use, might be worth commenting that or something?

tools/deploy.sh
```diff
@@ -0,0 +1,173 @@
...
+    echo "    Service is up. $START_TIME"
+else
+    echo "!! /meta/ did not return expected content. Response:"
+    echo "$META_RESPONSE" | head -5
+    echo "   The pod may still be starting — check again in a minute."
+    exit 1
+fi
+
+echo ""
+echo "==> Deploy complete: $TOOL_URL"
```
mahmoud, line 173: this is one of those scripts where if it works, it works, and if it doesn't we'll just fix it :P if there's a way to test it, it'll be pretty involved. I wonder if it makes sense to have a skill that runs it so that some automatic diagnosis and such can happen.

.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
mahmoud, line 1: nice. i assume it passes tox and your clicktesting?

package.json
```diff
@@ -0,0 +1,7 @@
+{
```
mahmoud, line 1: we want this at the root? not in the frontend dir?

requirements.txt
```diff
@@ -102,6 +88,11 @@ six==1.16.0
     # via
     #   html5lib
     #   python-dateutil
+setuptools==81.0.0
+    # via python-graph-core (pkg_resources namespace; implicit dep on Python 3.13)
+    # Pinned <82: setuptools 82 removed pkg_resources entirely; python-graph-core
+    # depends on it via namespace packages. Remove this pin once python-graph-core
+    # no longer uses pkg_resources (tracked in hatnote/montage#421).
```
mahmoud, line 95: good to know!

mahmoud (commented):
Big PR! First impressions: There are a few other things in the yaml that aren't secret per se? I wonder if that's too many env vars... and what's the deal with the huge .secrets.baseline ?

frontend/.env.production
lgelauff: try removing it - I recall that it keeps crashing if you do. Needs to be declared? I came across this earlier in #516

montage/public_endpoints.py
lgelauff: oauth is upgrading here from v1 to v2. This was because i had to register a new consumer after deleting the old credentials accidentally. 
Perhaps not the most lean implementation but made life easier.
Of course, if there is some reason to go back to v1, happy to register that instead.

montage/utils.py
lgelauff: OAuth 1.0a → 2.0 PKCE rewrite (413ea20)

montage/rdb.py
lgelauff: separate bug, that I think got in the way during testing. Was easier to fix than work around. 

Can cherrypick it as I can see there might be a conversation whether to hide paused (or completed) rounds now i think more about it.

tools/deploy.sh
lgelauff: yeah, personal preference, I guess :) Especially the smoke test was helpful to me, as it is explicit about when it's really ready.

.python-version
lgelauff: Good pointer, upgraded tox, and confirmed this passes. See #573

package.json
lgelauff: yes, see https://wikitech.wikimedia.org/wiki/Help:Toolforge/Building_container_images

lgelauff:
> Big PR! First impressions: There are a few other things in the yaml that aren't secret per se? I wonder if that's too many env vars... and what's the deal with the huge .secrets.baseline ?

Thanks for the comments and questions. In the new setup, we don't use a persistent file system any longer. It now also includes environment-specific variables. Moved it to env var. If  you want, we can make that distinction clearer and move them to the `USER_ENV_MAP`.  

The secrets baseline was for a pre-commit hook that I installed just in case Claude goes overboard and wants to commit something with a secret. The baseline are all known false positives. I realize now that I look more carefully that this can easily be condensed by skipping the test data.

lgelauff:
**Python version note:** This PR adds gunicorn 25.3.0, which requires Python >=3.10 — Python 3.9 is no longer supported as of this change. Minimum supported version is now Python 3.11 (local dev/CI); deployment target remains Python 3.13 (buildservice). Dropped `py39` from the tox matrix in ec20524.

lgelauff:
**Merge dependency:** `deployment.md` references `tools/migrate_prod_db.sql` (line 92), which is added by #505. This PR should not be merged before #505, otherwise the upgrade deployment instructions will reference a file that doesn't exist yet.

505 is now merged.

mahmoud (approved):
Updates + replies all seem good enough. This is still a big change, so deploy with care, and give some thought to how we might roll back. Start with dev, then beta, then prod, the practice will make perfect. :) Also, interesting that we're losing persistent filesystem; the main uses I had for it: 1) on-the-fly changes to workaround breakages (less of an issue as we get build/deploy more automated)  2) logs. Not clear to me how we're solving for logging with this.

lgelauff:
🤖 *This analysis was generated by Claude (claude-sonnet-4-6) and reviewed by Lodewijk.*

While reviewing the logging changes in this PR, I noticed one thing that disappears from the old `uwsgi.log` that isn't covered by any persistent log going forward: **client IP address**.

Here's how the coverage maps out:

| What uWSGI logged | Where it goes now |
|---|---|
| Method + path | `montage_api.log` (persistent NFS) + gunicorn stdout (ephemeral) |
| HTTP status code | `montage_api.log` (as "succeeded"/"raised exception") + gunicorn stdout |
| Response time | `montage_api.log` |
| Worker startup / lifecycle events | gunicorn stderr → `toolforge webservice buildservice logs` (ephemeral) |
| **Client IP** | **gunicorn stdout only — ephemeral, lost on pod restart** |

The gunicorn access log (`--access-logfile -`, added in this PR) does capture IPs, but only until the next restart. `montage_api.log` is richer in application terms (it has the Montage username and request params), but it never logs the IP.

Probably not worth blocking the PR on — IP logging isn't something Montage has actively used — but flagging it in case it matters for abuse tracking or audit purposes. If it does, the options are to write gunicorn's access log to an NFS path instead of stdout, or to add IP logging to `LoggingMiddleware` in `montage/mw/__init__.py`.

lgelauff:
> 🤖 _This analysis was generated by Claude (claude-sonnet-4-6) and reviewed by Lodewijk._

> Probably not worth blocking the PR on — IP logging isn't something Montage has actively used — but flagging it in case it matters for abuse tracking or audit purposes. If it does, the options are to write gunicorn's access log to an NFS path instead of stdout, or to add IP logging to `LoggingMiddleware` in `montage/mw/__init__.py`.

Looking more carefully at this... normally IP logging is carefully restricted in Wikimedia. For privacy reasons I prefer not to do this, not even within the pod lifetime. Given that users can only participate while logged in, and that there doesn't seem to be a real need, I am not implementing this at this time. But happy to discuss further if I'm missing something, @mahmoud !

lgelauff:
## Rollback runbook for the buildservice deploy

Captured from a two-reviewer pass on `tools/deploy.sh` + `deployment.md` (to be folded into `deployment.md` as a "Rolling back a bad deploy" section). The forward deploy has good pre-restart guardrails (SHA verify, port-8000 check, `/meta` smoke test) but no documented way *back*. This is that path.

### Principle
On the Toolforge buildservice, **a rollback is just a forward deploy of an older ref.** There is no reliable "promote a stored image" primitive to depend on — you rebuild the last-good commit and restart.

### Before every deploy — capture the anchor
Record, *before* building:
- the current **webservice type** (buildservice vs legacy) from `~/service.manifest`, and
- if buildservice, the current **running `RESULT_SHA`** (`toolforge build logs 2>&1 | grep RESULT_SHA | tail -1`).

That SHA is your rollback target. `deploy.sh` already computes it (line ~58) but discards it — worth persisting (see follow-up).

### Triggers
Cutover errors, `/meta` 500 / timeout, "no healthy upstream", or workers crash-looping after start.

### Procedure — by case
**A. Previous good state was buildservice at SHA `X`:**
```bash
toolforge build start https://github.com/hatnote/montage.git --ref <X>
# wait for Status: ok, confirm RESULT_SHA==X and gunicorn binds 0.0.0.0:8000
toolforge webservice buildservice restart --mount all
```
Rollback takes ~2–3 min (a rebuild), not seconds.

**B. Previous good state was the legacy python3.11 webservice:**
```bash
toolforge webservice stop
toolforge webservice start
```

### Database — no schema rollback needed
The `file_id` migration (`tools/migrate_prod_db.sql`) is **additive + nullable**, so old code tolerates the column. A **code** rollback does **not** require reverting the schema. `tools/revert_prod_db.sql` (`DROP COLUMN file_id`, idempotent) exists if ever truly needed, but the default is: **roll back code, leave the schema.**

### Roll back vs. fix forward
- **Fix forward** when the cause is understood and quick: a missing `file_id` column → run `migrate_prod_db.sql`; a blank/missing envvar → set it and restart.
- **Roll back** when the *image* is bad or won't bind port 8000, or the cause isn't obvious within a couple of minutes.

### Port-8000 timing (matters for rollback)
`deploy.sh` verifies the gunicorn port **before** restarting, so a port failure caught *pre-restart* means the old pod is still serving — **do nothing to the running service**, just rebuild a good ref. A failure only seen *post-restart* ("no healthy upstream") means you already cut over to the bad image and must roll back per Case A/B.

### Cross-tool safety
Every rollback command is account-scoped — **verify `whoami` shows the intended `tools.<env>` before each mutating command** (see #599). A misfire under the wrong `become` context is how the last incident happened.

### Suggested follow-ups (separate PRs)
- `deploy.sh`: persist the pre-deploy good SHA to a file, and print the rollback command in the smoke-test-failure branch instead of a dead-end `exit 1`.
- `deployment.md`: add this as a permanent "Rolling back a bad deploy" section + a "roll back vs fix forward" note.

PR #571, opened 2026-05-23
fix: juror round listing shows wrong rounds

lgelauff:
## Summary

- Fixes juror round listing to filter by round status, not just campaign status — finalized and cancelled rounds inside active campaigns were showing up and producing a confusing 400 \"round not active\" error when clicked
- Restricts the active view to rounds with `status IN (active, paused)` in the first pass, then narrows to `active` only (open question: should paused rounds be visible to jurors in read-only mode? see discussion)
- Fixes `get_index()` passthrough bug where `only_active=False` was ignored when calling `get_all_rounds_task_counts`, causing `/juror/campaigns/all` to always return an empty list
- Removes unused `router` import in `RoundEdit.vue`

Root cause: `e3069dd` (#350) added `finalize_round` which can finalize individual rounds without advancing the campaign, creating a state the juror listing never filtered for.

## Open question

Should paused rounds appear in the juror active view (read-only / greyed out) or be hidden entirely? Currently hidden (`== ACTIVE_STATUS`). Showing them (`IN (active, paused)`) matches how `Campaign.active_round` treats pause state, but jurors clicking a paused round get a 400 — that would need a frontend affordance or a dedicated endpoint.

## Test plan

- [ ] Juror index does not show finalized or cancelled rounds from active campaigns
- [ ] Juror index does not show paused rounds (current behaviour — revisit after discussion above)
- [ ] `/juror/campaigns/all` returns finalized rounds correctly
- [ ] `pytest montage/tests/` passes

### DIFF
--- a/frontend/src/components/Round/RoundEdit.vue
+++ b/frontend/src/components/Round/RoundEdit.vue
@@ -100,9 +100,6 @@ import Delete from 'vue-material-design-icons/Delete.vue'
 import alertService from '@/services/alertService'
 import adminService from '@/services/adminService'
 import dialogService from '@/services/dialogService'
-import { useRouter } from 'vue-router'
-
-const router = useRouter()
 const { t: $t } = useI18n()
 const props = defineProps({
   round: Object,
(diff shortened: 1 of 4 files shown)


lgelauff:
Extracted from #564

PR #598, opened 2026-07-04
feat:Added zoom entries on voting pages

### LINKED ISSUE #34: Full resolution view / zoom
Jurors should be able to zoom in to see a photo in full resolution. Jurors may want to inspect an image closely to see its quality.

Malavika05M:
**Description**
This PR implements an in-app image zoom and pan feature for single-image views (Yes/No, Rating, and Image Review Dialog).

Previously, the only way for a juror to inspect full-resolution details was to click "Show full-size," which ejected them to a new browser tab. It adds an explicit zoom mode by introducing a new component `ZoomImage.vue` that keeps the juror in the app, allows them to inspect details via scroll/pinch/drag, and dynamically loads higher resolution thumbnails as they zoom in.

**Related Issues**
Fixes #34 

**Screenshots**

https://github.com/user-attachments/assets/39f98f1f-c645-40e3-a442-ab38d15443aa


ItsAbhinavM:
@Malavika05M thank you for your contribution. I see that the exit icon during zoom kinda blends in with the entry, could you add some border or highlight to the icon in order to make it clearly visible. Thanks!

Malavika05M:
@ItsAbhinavM Fixed the issue in the latest commit.

montage-issue34

ItsAbhinavM (changes requested):
@Malavika05M I tried testing the changes and have some queries.
- Since the border color is white how does the close button look on lighter backgrounds
- Does the text `200% zoom level` change ? does it also cover scrolled zoom and the button click
- When you hover the close button it blends in with the image making it invisible.
- I can minimize the image as much I want on scrolling however on clicking the button it exits the zoom when you try to minimize beyond the image height.

Also I believe there is some issues with the formatting of the images. Even though it focuses on the keeping it at the center the resolution of image might be the factor causing this bug.
image

Malavika05M:
@ItsAbhinavM I kept the border color white permanently so that it could be more visible as you mentioned in the previous review. When I checked with a few of the images, it was working fine, but I'll look into it again. 

I'll check with the mentioned issues and send the resolved code asap

